PR #26, opened 2026-05-03
Update docs agent integration 14179451555119813426

aojea:


aojea:
/gemini can we extend the doc and the example to have another node exposing a custom mcp server and the local agent being able to query the exposed mcp server in the other node through the mesh?

PR #30, opened 2026-05-03
Add 'list' argument to mcp-client to list the available tool

kaisoz:
Listing available tools is a first-class feature of the MCP protocol. This PR adds a `-list` flag to `mcp-client` and makes it the default behavior


aojea:
Cool, thanks

PR #50, opened 2026-05-07
Allow discovering remote services by type

kaisoz:


cmd/sam-node/gate.go
```diff
@@ -134,6 +135,34 @@ func (n *SamNode) HandleMCPStream(s network.Stream) {
...
+		respData, err := json.Marshal(services)
+		if err != nil {
+			return nil, nil, err
+		}
+		return &mcp.CallToolResult{
+			Content: []mcp.Content{
+				&mcp.TextContent{Text: string(respData)},
+			},
+		}, nil, nil
+	})
```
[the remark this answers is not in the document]
aojea: @kaisoz there should be only one `list_local_services`, no?

cmd/sam-node/node.go
```diff
@@ -974,65 +994,152 @@ func (n *SamNode) IsServiceRegistered(serviceName string) bool {
...
+				logger.Warnf("[Discovery] catalog fetch from %s failed: %v", peerID, err)
+				return
+			}
+			results <- peerCatalog{peerID: peerID, services: services}
+		}(p.ID)
+	}
+	wg.Wait()
+	close(results)
+
+discovered := []*api.DiscoveredProvider{}
```
aojea: is this format correct?

cmd/sam-node/node.go
```diff
@@ -974,65 +994,152 @@ func (n *SamNode) IsServiceRegistered(serviceName string) bool {
...
+			services, err := n.fetchRemoteServiceCatalog(fanoutCtx, peerID, typeStr)
+			if err != nil {
+				logger.Warnf("[Discovery] catalog fetch from %s failed: %v", peerID, err)
+				return
+			}
+			results <- peerCatalog{peerID: peerID, services: services}
+		}(p.ID)
+	}
+	wg.Wait()
+	close(results)
```
aojea: do we close the result channel before draining it?

aojea:
overall good
-  one question about duplication of methods on the mcp server
- other about the gofmt, or indentantion
- other about the possibility of goruotines deadlocks or leaks, draining channels and using the sync waitgroups need to be carefully handled

cmd/sam-node/gate.go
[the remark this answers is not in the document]
kaisoz: Agree (hence the TODO). I didn't do it before to avoid to have a big PR but is absolutely needed. I moved the MCP tools implementation to the `mcp_handlers.go` and refactored both `mcp.go` and `gate.go`. PTAL!

cmd/sam-node/node.go
kaisoz: 🤦🏻 This was a Gemini commit. Fixed

cmd/sam-node/node.go
kaisoz: Yes.. and that works because the channel keeps the values when closed. However, this makes the main thread wait till the routines have finished, breaking the streaming. Fixed!

PR #64, opened 2026-05-22
find_remote_tools: discover hosted MCP service tools across the mesh

kaisoz:


cmd/sam-node/aggregator.go
```diff
@@ -0,0 +1,92 @@
...
+// Best-effort: any failure leaves manifest fields nil/empty and returns
+// nil so RegisterService is not aborted. The service remains reachable
+// via the existing libp2p-HTTP proxy regardless.
+func (n *SamNode) aggregateServiceTools(ctx context.Context, manifest *ServiceManifest) error {
+	if manifest == nil || manifest.Info == nil || manifest.Handler == nil {
+		return nil
+	}
+	serviceName := manifest.Info.Name
+
+	loopback := httptest.NewServer(manifest.Handler)
```
aojea, line 32: /gemini we must not use test libraries on production code , also why do we need a loopback server? we already have one server, can't we use just a new. handler and reuse the existing server?
I'm unclear on why we need this if we just want to assemble a manifest, do we need all this mcp dependencies?

aojea: /gemini is it really necessary? since we just want to aggregate all this information meanwhile we just populate the sdk type we should be fine, why do we need to create the additional overhead of clienting a fake server and client? can we make this simpler and leaner?

aojea: /gemini I don't like this approach, sounds very hacky, can we have another simpler alternative

aojea: /gemini the point is that we should not need to use a loopback server, this is just to put the existing information in a determined type, we don't need the overhead just to put data into some fields

aojea: /gemini we don't need to use mcp.StreamableClientTransport or mcp.Client, qw just need to create the ServiceManifest

aojea: /gemini I do not care, we are not going to use this pattern , so look for another solution that does not require loopback servers and keeping state in the background vendoring the entire sdk

aojea: /gemini help outline the structure for such a DirectTransport implementation

kaisoz:
As discussed offline, this PR will be closed in favour of a refactor which extracts the services logic to their own entity. This PR has been superseeded by #65

PR #97, opened 2026-06-05
docs: add SAM agent skill

### LINKED ISSUE #62: Create sam SKILL so agents can know how to use it

AsyncAssassin:
## Summary

- Add a root-level `SKILL.md` for agents working with SAM (Sovereign Agent Mesh).
- Document the local `sam-node` MCP endpoint and the mesh inspection, service discovery, tool discovery, describe, and call workflow.
- Include safety guidance for non-loopback endpoints, sensitive data, side-effecting tools, and AWS SAM ambiguity.

## Testing

- Parsed the `SKILL.md` frontmatter.
- Ran the documented MCP workflow against a local Docker SAM mesh with a remote calculator MCP service.
- Include safety guidance for non-loopback endpoints, sensitive data, and side-effecting tools, and explicitly exclude AWS SAM / Serverless Application Model tasks.

Closes #62.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -1,0 +1,110 @@
+---
+name: sam
+description: "Use this skill only for SAM (Sovereign Agent Mesh) tasks through a local sam-node MCP server: inspecting mesh state, discovering reachable services/tools, describing namespaced remote MCP tools, and calling them with JSON-object arguments. Do not use for AWS SAM or Serverless Application Model tasks."
+---
+
+# SAM Agent Skill
+
+Use SAM when a local `sam-node` is available and the task needs capabilities
+hosted by peers in a SAM mesh. Prefer local tools first. Reach into the mesh
+only when the needed capability is not available locally.
+
+## Connect To The Local Node
+
+Connect to the local `sam-node` MCP SSE endpoint:
+
+```text
+http://127.0.0.1:8080/mcp/events
+```
+
+If the user or local configuration explicitly provides a trusted non-loopback
+endpoint, ask the user to confirm the exact endpoint and trust boundary before
+connecting. Prefer loopback endpoints (`127.0.0.1`, `localhost`, or `::1`), and
+do not send secrets or sensitive data to non-loopback endpoints unless the user
+explicitly approves the destination and data. Do not probe, guess, or scan
+non-local bind addresses. The server also registers `/mcp/message` as the
+paired MCP message endpoint; MCP clients and SDKs normally handle this after
+connecting through `/mcp/events`.
+
+## Inspect The Mesh
+
+Start by understanding the local node and mesh state:
+
+- Use `get_mesh_info` with `{}` to inspect `connected_peers`, `dht_size`, and
+  `hub_peer_id`.
+- Use `list_local_services` with `{}` to see services registered on the local
+  node.
+
+## Discover Remote Capabilities
+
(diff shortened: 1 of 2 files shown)


SKILL.md
```diff
@@ -0,0 +1,110 @@
...
+- Narrow `find_remote_tools` with `service_name` or `peer_id` when you already
+  know the target.
+- Mesh-wide searches fetch each peer's catalog on a best-effort basis and may
+  return an empty array when no reachable aggregated tools are found. Discovery
+  failures or explicit `peer_id` lookup failures are returned as errors.
+
+## Describe Before Calling
+
+For namespaced tools returned by `find_remote_tools`, call
+`describe_remote_tool` before `call_remote_tool`.
```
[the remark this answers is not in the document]
AsyncAssassin: Updated the wording to clarify that all tools returned by `find_remote_tools` are namespaced and must be described before calling.

kaisoz:
Since this is a skill to be **delivered** to agents, let's move it to `agents/skills`. The `agents` folder can contain all types of documents intended to be consumed by agents (e.g. prompts)

SKILL.md
```diff
@@ -0,0 +1,110 @@
...
+description: "Use this skill only for SAM (Sovereign Agent Mesh) tasks through a local sam-node MCP server: inspecting mesh state, discovering reachable services/tools, describing namespaced remote MCP tools, and calling them with JSON-object arguments. Do not use for AWS SAM or Serverless Application Model tasks."
+---
+
+# SAM Agent Skill
+
+Use SAM when a local `sam-node` is available and the task needs capabilities
+hosted by peers in a SAM mesh. Prefer local tools first. Reach into the mesh
+only when the needed capability is not available locally.
+
+## Connect To The Local Node
```
kaisoz: Drop this section. The model has no capabilities to configure the harness (it'd need a tool for that). The skill should assume that the SAM node tools are already callable and only cover what to do with them

SKILL.md
```diff
@@ -0,0 +1,110 @@
+---
+name: sam
+description: "Use this skill only for SAM (Sovereign Agent Mesh) tasks through a local sam-node MCP server: inspecting mesh state, discovering reachable services/tools, describing namespaced remote MCP tools, and calling them with JSON-object arguments. Do not use for AWS SAM or Serverless Application Model tasks."
```
kaisoz: I don't think this description is accurate enough. For an agent to choose the skill, the task would already have to be framed as "a SAM task", but an agent that doesn't yet know it should reach into a mesh won't phrase it that way. 

Let's make the description trigger on intent, not the name "SAM", something like

```
Use when a task needs a tool or capability that isn't available locally and a sam-node MCP server is connected — reach remote agents on a SAM (Sovereign A gent Mesh) network: inspect mesh state, discover reachable services, describe namespaced remote MCP tools, and call them. Not for AWS SAM / Serverless Application Model
```
WDYT?

AsyncAssassin:
Thanks @kaisoz, updated it.

The skill now lives in `agents/skills/sam/SKILL.md`. I also removed the local MCP setup section and changed the description to focus on the “local tools are not enough” case, instead of requiring the task to already be described as a SAM task.

kaisoz (approved):

kaisoz:
lgtm, we can start from her and iterate testing with different harnesses

Assigning @aojea for the final pass and merge

PR #170, opened 2026-06-30
Service catalog: announce foundation, external catalog, consumer-side resolver

kaisoz:
## Summary
Adds a **service catalog** that accelerates mesh discovery, in three layers:

1. **Announce foundation (node-side):** every node publishes a signed `ServiceAnnounce` for its local services on the gossip topic `/sam/service/announce/v1`, on the existing reprovide tick. Self-signed (verified via the announcer's `peer_id`), no trusted-key list.
2. **External catalog (`cmd/sam-catalog`):** a standalone process that connects to a node's sidecar, bootstraps via `/sam/service/discover`, tails a new binary-safe `/sam/service/announce/stream` SSE endpoint, maintains an in-memory store with TTL sweep + periodic re-walk, serves a `query_catalog` MCP tool, and self-registers as a `CATALOG` service.
3. **Consumer-side resolver (node-side):** discovery resolves via a reachable catalog first (transport-agnostic endpoint resolution — a catalog this node hosts is queried directly over HTTP; a remote one over libp2p), falling back to the DHT on miss/error/empty.

## Tests
- Unit tests for announce sign/verify, the catalog store (TTL), the queryCatalog client + entry→provider mapping.
- Integration tests (mock hub): announce propagation, catalog end-to-end, discovery + DHT fallback. (The catalog-accelerated path over biscuit-authed libp2p is skipped under the mock hub, which can't issue real biscuits.)
- New `tests/e2e/catalog.bats` (real hub): three cases — retrieval via catalog, on-the-fly service pickup, DHT fallback.

## Notes
- DHT stays source-of-truth; the catalog is an accelerator (empty/unreachable → DHT fallback).
- `cmd/sam-catalog`'s `--own-url` defaults to the bind address; set a node-reachable address when binding to a non-loopback interface.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ build:
 	go build -v -o "$(OUT_DIR)/sam-node" ./cmd/sam-node
 	go build -v -o "$(OUT_DIR)/sam-hub" ./cmd/sam-hub
 	go build -v -o "$(OUT_DIR)/mcp-client" ./cmd/mcp-client
+	go build -v -o "$(OUT_DIR)/sam-catalog" ./cmd/sam-catalog
 
 .PHONY: proto
 proto:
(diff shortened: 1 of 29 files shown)


aojea:
this is a very large refactor on an area that has changed multiple times so I do not think we are able to rebase it, let's tackle this in a different PR

PR #188, opened 2026-07-09
feat(node): Add the `--join` flag to the `run` command to register and enroll in one step

kaisoz:


aojea:
rebased and merged in https://github.com/google/sam/pull/265

PR #224, opened 2026-07-23
feat(security): transition from static policies.yaml to dynamic relat…

aojea:
…ional SQL policy API

- Replaced local file-based policies.yaml with centralized SQL tables (roles and bindings) in Postgres/SQLite on the Control Plane.
- Exposed a REST API (/policies) to dynamically retrieve and update policies on the Hub.
- Refactored the policy compiler to build Biscuit Datalog rules from relational DB snapshots locally on the node.
- Added support for periodic policy sync interval configuration flag (--policy-sync-interval) on the node.
- Gracefully handled OIDC callback local server shutdown to prevent TCP reset issues during enrollment.
- Removed obsolete --policy-file flags, volumes, and ConfigMaps from all GKE template and kind development configurations.
- Integrated automated policy seeding via background port-forward curl commands post-rollout in GitHub actions and local kind runner scripts.
- Updated policy and control plane website guides to document the new JSON API and database model.
- Fixed all golangci-lint issues across the codebase.
- Verified that all unit, integration, and E2E test suites pass successfully.


aojea:
/gemini review and update the comments that are stale

aojea:
/gemini review again and close comments that are no longer relevant

PR #249, opened 2026-08-09
Fix router re-enroll process

kaisoz:
The router reads its OIDC token once at startup and caches it in `r.config.OIDCToken`. Re-enrollment reuses that cached value, so if the token is rotated on disk (like with a projected service account token), the router replays an expired JWT and can never re-enroll.

`reEnroll` now re-reads the token from `--jwt-path` before using it, which is what the node already does on its renewal path (`internal/node/node.go`). The bootstrap-token path is unchanged.

### DIFF
--- a/internal/router/router.go
+++ b/internal/router/router.go
@@ -148,35 +148,11 @@ func (r *Router) Start() error {
 		return err
 	}
 
-	if r.config.BootstrapToken == "" && r.config.BootstrapTokenPath != "" {
-		tokenData, err := os.ReadFile(r.config.BootstrapTokenPath)
-		if err != nil {
-			return fmt.Errorf("failed to read bootstrap token from path %s: %w", r.config.BootstrapTokenPath, err)
-		}
-		r.config.BootstrapToken = strings.TrimSpace(string(tokenData))
-	}
-
-	if r.config.BootstrapToken != "" {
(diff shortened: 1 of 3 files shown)


internal/router/router.go
aojea: See the enroll logic, that is the one we need to use for reenroll, worth considering consolidating the code to avoid drift and call from both places

internal/router/router.go
```diff
@@ -481,7 +494,11 @@ func (r *Router) reEnroll() error {
 
 	if r.config.BootstrapToken != "" {
 		return r.enrollBootstrap(r.Host.ID())
-	} else if r.config.OIDCToken != "" {
+	}
```
aojea: it is missing `r.config.BootstrapTokenPath != ""`  , see https://github.com/google/sam/pull/249/changes#r3747731410

internal/router/router_test.go
```diff
@@ -684,3 +685,38 @@ func TestRouterGossipSubBannedEvent(t *testing.T) {
 		t.Errorf("expected peer %s to be stored in bannedPeers blocklist upon receiving MeshEvent_BANNED", bannedPeerIDStr)
 	}
 }
+
```
aojea, line 689: we need to test for all possible combinations, reenroll with bootstrap token, reenoll, with bootstraptokenpath, oidc token, oidc token path

PR #279, opened 2026-08-17
Improve SAM kind dev environment

kaisoz:
- Expose the kind mesh through Gateway API LoadBalancer addresses using cloud-provider-kind, dropping the NodePort lane and all port-forwards.
- Announce the router at its own node IP over TCP and QUIC, and keep it off workers that host sam-nodes.
- Serve the console under `/console` on the mesh gateway address via a new `--base-path`.
- Declare the dev mesh layout in `mesh-config.yaml`, with example services deployed as sidecars.
- Make `sam:role:node`'s allowed services configurable instead of a hardcoded list of e2e test fixtures.

Plus: the local node now enrolls with a bootstrap token minted through `/admin`, the bundled Dex test
account is no longer called admin, the console waits for its issuer, the `greeter-mcp` example works
again, the docs stop describing the kind mesh in terms of the cloud deployment, and `run.sh` no longer
leaks envoy containers or hangs on Dex discovery.


internal/console/server.go
```diff
@@ -119,12 +121,19 @@ func NewServer(cfg Config) (*Server, error) {
...
+		routes.HandleFunc("/auth/session", s.HandleSession)
+	}
+	routes.HandleFunc("/auth/logout", s.HandleLogout)
+	routes.HandleFunc("/info", s.HandleInfo)
+
+	// Serve under BasePath too, so the links this server emits resolve without the proxy
+	// stripping the prefix. Still served at the root for proxies that do strip it.
+	if s.cfg.BasePath != "" {
+		s.mux.Handle(s.cfg.BasePath+"/", http.StripPrefix(s.cfg.BasePath, routes))
 	}
```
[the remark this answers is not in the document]
kaisoz: Nothing in the current codebase can trigger it, this requires a malformed `--base-path`. We could treat is as a follow-up

tests/e2e/lib/container_mesh.bash
```diff
@@ -282,7 +286,9 @@ if [[ -z "${MESH_HELPERS_LOADED:-}" ]]; then
       --set controlPlane.replicaCount=2 \
       --set controlPlane.hostPort=8080 \
       --set router.useOidcToken=false \
-      --set router.hostPort=4501
+      --set router.hostPort=4501 \
+      --set 'router.externalAddrs={/dns4/sam-router/tcp/4501}' \
```
aojea, line 290: why do we need to define this now? before it was able to work without it

charts/sam-mesh/templates/dex-deployment.yaml
```diff
@@ -1,3 +1,38 @@
+{{/*
+Dex config body. Named so the Deployment can checksum it: nothing else would restart Dex
+when the issuer or the redirect URIs change.
+*/}}
+{{- define "sam-mesh.dexConfig" -}}
```
aojea, line 5: I do not know .... this seems violating multiple layers, we deploy dex for simplicity and having a standalone deployment ... but theoretically dex should be an indpendent component, leaking the config to the chart makes this more tighly coupled

charts/sam-mesh/templates/console-deployment.yaml
```diff
@@ -24,10 +28,27 @@ spec:
...
+          # just any control plane.
+          agreed=0
+          until [ "$agreed" -ge 3 ]; do
+            if curl -sf "${CP_URL}/info" | grep -qF '{{ $issuer }}'; then
+              agreed=$((agreed+1))
+            else
+              echo "Control plane not serving issuer {{ $issuer }} yet, sleeping..."
+              agreed=0
+            fi
+            sleep 2
```
aojea, line 44: can you explain a bit more what problem is this solving? what does it means  "old issuer" ?

kaisoz:
Discussed offline with @aojea. Since we are approaching a new cut, I think is better to close this PR and focus on making the chart production ready, which will lead to improvements in the kind dev environment

PR #280, opened 2026-08-17
fix(site): restore scrolling so the landing demo stays reachable

roushou:
## What

Hello! This is a fix to the landing page demo being cut off at the bottom and unreachable.
The bottom of the demo was clipped out due to `overflow: hidden` on the body even on scroll.

## Changes

- `overflow: hidden` → `overflow-x: hidden` — restores vertical scrolling
- Demo card styles moved from inline HTML into the landing SCSS

| Before | After (scrolled down) |
| :---: | :---: |
| image | image |

### DIFF
--- a/site/assets/scss/_styles_project.scss
+++ b/site/assets/scss/_styles_project.scss
@@ -31,7 +31,8 @@
   justify-content: center;
   align-items: center;
   position: relative;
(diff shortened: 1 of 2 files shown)


aojea:
Thanks

PR #293, opened 2026-08-22
fix(api): sort label keys in ValidateLabels for deterministic errors

HosniBelfeki:
## Summary

- `api.ValidateLabels` iterates the input `map[string]string` directly and returns on the first invalid key it visits. Because Go map iteration is randomized, the reported key — and therefore the diagnostic surface operators see — varies between runs. Two reproductions of the same bad config could print different keys, and a flaky unit test against a multi-bad-key input would be unreliable.
- This change collects the keys into a slice, sorts them with `sort.Strings`, and walks them in that order. The reported key for any given input is now deterministic (the lexicographically smallest invalid one), and the function stays a no-op for valid input. The pattern mirrors the existing one already used for `LabelsToFacts` in `api/datalog.go`, so no new abstraction is introduced.
- Two regression tests are added directly next to the change in `api/labels_test.go`:
  - `TestValidateLabels_ReportsFirstErrorDeterministically` calls `ValidateLabels` 200 times on a five-bad-key input and asserts the error always quotes the lexicographically smallest invalid key. The previous nondeterministic implementation would have leaked a different key on different iterations.
  - `TestValidateLabels_AllKeysValidStaysSorted` asserts the happy path is unaffected by the sort.

No new module is added to `go.mod`; only the stdlib `sort` package is used, in line with the repository's dependency policy.

## Tests

```
$ go test -v -run TestValidateLabels ./api/...
=== RUN   TestValidateLabels
--- PASS: TestValidateLabels (0.00s)
=== RUN   TestValidateLabels_ReportsFirstErrorDeterministically
--- PASS: TestValidateLabels_ReportsFirstErrorDeterministically (0.00s)
=== RUN   TestValidateLabels_AllKeysValidStaysSorted
--- PASS: TestValidateLabels_AllKeysValidStaysSorted (0.00s)
PASS
ok      github.com/google/sam/api  0.836s

$ go test -count=1 ./api/...
ok      github.com/google/sam/api  0.856s

$ go vet ./api/...
(no findings)

$ gofmt -l api/labels.go api/labels_test.go
(clean)
```

The package-local test result above is the only check I can run locally without gcc/CGO-enabled race; CI on `google/sam` will run the full gated matrix on top of this.

### DIFF
--- a/api/labels.go
+++ b/api/labels.go
@@ -17,6 +17,7 @@ package api
 import (
 	"fmt"
 	"regexp"
+	"sort"
 	"strings"
 )
 
@@ -63,13 +64,23 @@ func ValidateLabelValue(value string) error {
 	return nil
 }
 
-// ValidateLabels checks every key and value in a label set.
+// ValidateLabels checks every key and value in a label set. Keys are visited
+// in lexicographic order, so the error returned for a given input is
+// deterministic and stable across runs (Go's map iteration is randomized).
 func ValidateLabels(labels map[string]string) error {
-	for k, v := range labels {
+	if len(labels) == 0 {
+		return nil
+	}
+	keys := make([]string, 0, len(labels))
+	for k := range labels {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, k := range keys {
 		if err := ValidateLabelKey(k); err != nil {
 			return err
 		}
-		if err := ValidateLabelValue(v); err != nil {
+		if err := ValidateLabelValue(labels[k]); err != nil {
 			return err
 		}
 	}
--- a/api/labels_test.go
+++ b/api/labels_test.go
@@ -65,3 +65,45 @@ func TestValidateLabels(t *testing.T) {
 		t.Error("ValidateLabels(bad key): expected error, got nil")
 	}
 }
+
+// TestValidateLabels_ReportsFirstErrorDeterministically guards against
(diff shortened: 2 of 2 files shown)


HosniBelfeki:
Addressed the two Gemini Code Assist suggestions in the latest force-push (`e5a8338`):

- `ValidateLabels` now short-circuits when the input map is empty or nil, so the common "no labels" case skips the slice allocation entirely.
- The determinism regression test now exercises a mix of valid keys and three independently invalid keys (`a!key`, `b_ bad`, `c,key`). It still asserts the reported error always quotes the lexicographically smallest invalid key, which is `"a!key"` in the new input — so the assertion only passes if sort order drives the choice, not luck.

Local validations re-run after the change:

```
$ go test -v -count=1 -run TestValidateLabels ./api/...
=== RUN   TestValidateLabels
--- PASS: TestValidateLabels (0.00s)
=== RUN   TestValidateLabels_ReportsFirstErrorDeterministically
--- PASS: TestValidateLabels_ReportsFirstErrorDeterministically (0.00s)
=== RUN   TestValidateLabels_AllKeysValidStaysSorted
--- PASS: TestValidateLabels_AllKeysValidStaysSorted (0.00s)
PASS
ok      github.com/google/sam/api  0.776s

$ go vet ./api/...
(no findings)

$ gofmt -l api/labels.go api/labels_test.go
(clean)
```

aojea:
Thanks

PR #301, opened 2026-08-24
use GET for enroll status polling #297

### LINKED ISSUE #297: Bootstrap enrollment cannot complete: clients POST to /enroll/status but the server only accepts GET (405)
## Summary

`sam-node` and `sam-router` poll enrollment status with **POST**, but `HandleEnrollStatus` only accepts **GET** with a `peer_id` query parameter. Every poll returns `405 Method Not Allowed`, so a node that has been approved by an admin can never retrieve its Biscuit.

The effect is that **bootstrap-token enrollment cannot complete**. This is the headless path (CI, containers, non-interactive deployments), so it is not reachable via the interactive OIDC flow.

Confirmed present at `674af93e8942fee8f05f6f9aec7d210fed6e947b` (current `main` at time of filing).

## The mismatch

Server — `internal/controlplane/server.go`, `HandleEnrollStatus`:

```go
if r.Method != http.MethodGet {
    http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
    return
}
peerID := r.URL.Query().Get("peer_id")
```

Client — `internal/node/enroll.go`:

```go
statusURL := controlPlaneURL + "/enroll/status"
hReq, err := http.NewRequestWithContext(ctx, "POST", statusURL, bytes.NewReader(statusData))
hReq.Header.Set("Content-Type", "application/x-protobuf")
```

Client — `internal/router/router.go`:

```go
statusResp, err := client.Post(r.config.ControlPlaneURL+"/enroll/status",
                               "application/x-protobuf", bytes.NewReader(statusData))
```

## Reproduction

```bash
# 1. Control plane (manual approval is the default)
sam-control-plane \
  --bind-address 127.0.0.1:8090 \
  --issuer http://127.0.0.1:5570 \
  --allowed-audiences my-audience \
  --admin-token-path ./admin-token \
  --db-dsn ./cp.db

# 2. Mint a node bootstrap token
curl -s -X POST http://127.0.0.1:8090/admin/bootstrap-tokens \
  -H "Authorization: Bearer $(cat admin-token)" \
  -H "Content-Type: application/json" \
  -d '{"role":"sam:role:node","ttl_hours":2,"max_usages":5}'

# 3. Start a node with that token
sam-node run --control-plane http://127.0.0.1:8090 \
  --bootstrap-token sam-bt-... --allow-loopback

# 4. Approve the pending enrollment
curl -s -X POST http://127.0.0.1:8090/admin/enrollments/<id>/approve \
  -H "Authorization: Bearer $(cat admin-token)"
# -> 200 Enrollment approved
```

The node never picks the approval up:

```
WARN sam-node enroll.go:281 Status check returned status 405 Method Not Allowed: Method not allowed
WARN sam-node enroll.go:281 Status check returned status 405 Method Not Allowed: Method not allowed
...
```

`sam-router` behaves identically (`router.go:408`).

The endpoint itself works — only the verb is wrong:

```bash
curl -i "http://127.0.0.1:8090/enroll/status?peer_id=12D3KooW..."
# HTTP/1.1 200 OK
# Content-Type: application/x-protobuf   (589 bytes: the approved Biscuit)

curl -i -X POST "http://127.0.0.1:8090/enroll/status"
# HTTP/1.1 405 Method Not Allowed
```

## Suggested fix

Changing the two clients to GET is the smaller change, since the server already reads `peer_id` from the query string and the request body is otherwise unused:

```go
// internal/node/enroll.go
statusURL := controlPlaneURL + "/enroll/status?peer_id=" + url.QueryEscape(n.Host.ID().String())
hReq, err := http.NewRequestWithContext(ctx, "GET", statusURL, nil)
```

```go
// internal/router/router.go
statusResp, err := client.Get(r.config.ControlPlaneURL + "/enroll/status?peer_id=" + peerID.String())
```

`statusData`/`statusReq` then become unused in both call sites and can be removed.

Alternatively, have `HandleEnrollStatus` accept POST with the existing protobuf body — that keeps the wire format but is a larger server-side change.

I verified the client-side fix locally: after patching, the router logged `Router bootstrap enrollment approved! Biscuit received.` and three nodes enrolled and formed a working mesh.

## Notes

- A test covering approve-then-collect over HTTP would catch this class of bug; the current tests appear to exercise the handlers directly rather than through the client polling path.
- Happy to open a PR if that is useful — I would need to sort out the CLA first.

Found while building an external integration against SAM. Environment: Windows, Go 1.25.14, binaries built from source.

IbrahimAhmed8:
clients used POST instead of GET for `/enroll/status` causing 405s.

* switched clients to use GET
* put peer_id in the query string
* removed unused payloads and headers
* renamed a url var to fix shadowing

fixes #297

### DIFF
--- a/api/sam.pb.go
+++ b/api/sam.pb.go
@@ -541,50 +541,6 @@ func (x *EnrollResponse) GetExpiration() int64 {
 	return 0
 }
 
-type EnrollmentStatusRequest struct {
-	state         protoimpl.MessageState `protogen:"open.v1"`
-	PeerId        string                 `protobuf:"bytes,1,opt,name=peer_id,json=peerId,proto3" json:"peer_id,omitempty"`
-	unknownFields protoimpl.UnknownFields
-	sizeCache     protoimpl.SizeCache
-}
-
-func (x *EnrollmentStatusRequest) Reset() {
-	*x = EnrollmentStatusRequest{}
-	mi := &file_api_sam_proto_msgTypes[5]
-	ms := protoimpl.X.MessageStateOf(protoimpl.Pointer(x))
-	ms.StoreMessageInfo(mi)
-}
-
-func (x *EnrollmentStatusRequest) String() string {
-	return protoimpl.X.MessageStringOf(x)
-}
-
-func (*EnrollmentStatusRequest) ProtoMessage() {}
-
-func (x *EnrollmentStatusRequest) ProtoReflect() protoreflect.Message {
-	mi := &file_api_sam_proto_msgTypes[5]
-	if x != nil {
-		ms := protoimpl.X.MessageStateOf(protoimpl.Pointer(x))
-		if ms.LoadMessageInfo() == nil {
-			ms.StoreMessageInfo(mi)
-		}
-		return ms
-	}
-	return mi.MessageOf(x)
-}
-
-// Deprecated: Use EnrollmentStatusRequest.ProtoReflect.Descriptor instead.
-func (*EnrollmentStatusRequest) Descriptor() ([]byte, []int) {
-	return file_api_sam_proto_rawDescGZIP(), []int{5}
-}
-
-func (x *EnrollmentStatusRequest) GetPeerId() string {
-	if x != nil {
-		return x.PeerId
-	}
-	return ""
-}
-
 type BootstrapEnrollRequest struct {
 	state          protoimpl.MessageState `protogen:"open.v1"`
 	BootstrapToken string                 `protobuf:"bytes,1,opt,name=bootstrap_token,json=bootstrapToken,proto3" json:"bootstrap_token,omitempty"`
@@ -600,7 +556,7 @@ type BootstrapEnrollRequest struct {
 
 func (x *BootstrapEnrollRequest) Reset() {
 	*x = BootstrapEnrollRequest{}
-	mi := &file_api_sam_proto_msgTypes[6]
+	mi := &file_api_sam_proto_msgTypes[5]
 	ms := protoimpl.X.MessageStateOf(protoimpl.Pointer(x))
 	ms.StoreMessageInfo(mi)
 }
@@ -612,7 +568,7 @@ func (x *BootstrapEnrollRequest) String() string {
 func (*BootstrapEnrollRequest) ProtoMessage() {}
 
 func (x *BootstrapEnrollRequest) ProtoReflect() protoreflect.Message {
-	mi := &file_api_sam_proto_msgTypes[6]
+	mi := &file_api_sam_proto_msgTypes[5]
 	if x != nil {
 		ms := protoimpl.X.MessageStateOf(protoimpl.Pointer(x))
 		if ms.LoadMessageInfo() == nil {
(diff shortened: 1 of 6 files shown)


aojea:
Things I specifically checked and am happy with:

- Renaming the `url` local to `enrollURL`. Good catch.
- Switching the router to `http.NewRequestWithContext(r.ctx, ...)` is an
  improvement: the previous `client.Post` ignored cancellation entirely
  and would hang for the full 30s client timeout on shutdown.
- The Gateway route is `type: Exact, value: /enroll/status`
  (.github/k8s/sam-control-plane-template.yaml:215). 

Three changes before we merge

1. This needs a test

A verb mismatch between two components belongs at the integration
level, in `tests/integration/`: start a control plane with
`AutoApproveEnrollment = false`, run the real client polling path, approve through the
admin API, and assert the client ends up holding a Biscuit.

We already have a test that calls the handler is what we already have, and it is what missed this.

2. `router.go`: the new error branch never terminates

```
    req, err := http.NewRequestWithContext(...)
    if err != nil {
        logger.Warnf("failed to create enrollment status request: %v", err)
        continue
    }
```

`NewRequestWithContext` fails only on a malformed method or URL, which is
deterministic — so a bad `ControlPlaneURL` now logs a warning every 2s forever with no
exit path. `enroll.go` returns an error in the identical situation.

The URL is loop-invariant in both files. Build it once above the `for`/`select` and
return the error, which fixes the spin and removes the duplicated per-tick work:

    statusURL := r.config.ControlPlaneURL + "/enroll/status?peer_id=" + url.QueryEscape(peerID.String())
    for {
        select {
        ...
        case <-ticker.C:
            req, err := http.NewRequestWithContext(r.ctx, http.MethodGet, statusURL, nil)
            if err != nil {
                return fmt.Errorf("failed to create status request: %w", err)
            }

Please make `enroll.go` symmetric. While you're there, use `http.MethodGet` rather
than the `"GET"` string literal in both files.

3. `api.EnrollmentStatusRequest` seems now dead. Check if we no longer need and we can delete it. Just check you recreate the protos

IbrahimAhmed8:
thanks @aojea. I pushed the changes:

- moved the url build out of the loops and used http.MethodGet
- router.go error handling now matches enroll.go
- dropped the dead EnrollmentStatusRequest and regenerated protos
- added the integration test

i also set the protoc version string to v3.21.12 so the verify passes.
it looks like the bats and e2e android jobs are failing to start because of a github billing limit on this repo

aojea:
> it looks like the bats and e2e android jobs are failing to start because of a github billing limit on this repo

@IbrahimAhmed8 can you please rebase to fix that?

IbrahimAhmed8:
I rebased on main and applied the bot suggestions. actually pushed an empty commit to retrigger the bats job before realizing it forces a full run of the whole suite. sorry for burning the actions minutes especially with the billing limit issue you mentioned

aojea:
CI fixed in https://github.com/google/sam/pull/312
Please rebase.
Apologies for the inconvenience

aojea:
Two small things and then I'll merge:

1. Add `w.Header().Set("Cache-Control", "no-store")` to `writeEnrollResponse`
   (internal/controlplane/server.go:1048). Now that this is a GET returning a Biscuit,
   the response is cacheable ( it wasn't as a POST).
2. Please squash into a single commit

IbrahimAhmed8:
i looked at the raw logs and its failing on test 23 in relay.bats line 91 where it reports 0 peers. looks like a dht flake

aojea:
> i looked at the raw logs and its failing on test 23 in relay.bats line 91 where it reports 0 peers. looks like a dht flake

yeah, unrelated flake, we need to deflake that, but is not related to this change

IbrahimAhmed8:
> > i looked at the raw logs and its failing on test 23 in relay.bats line 91 where it reports 0 peers. looks like a dht flake
> 
> yeah, unrelated flake, we need to deflake that, but is not related to this change

i added the cache header and combined everything into one commit

aojea:
Thanks

PR #319, opened 2026-08-26
refactor: extract tool fetch loop body for safe cleanup

IbrahimAhmed8:
In internal/node/mcp_handlers.go, fetchRemoteToolCatalogue now ensures cleanup occurs per iteration.

The previous implementation was functionally correct but introduced a latent risk. If session.ListTools panics or if a conditional return is introduced later, active streams and sessions would leak

Changes:
Extracted the loop body into a helper method named fetchToolsForRemoteService. This scopes the defer statement to the iteration scope rather than the parent function scope.

A standard defer inside the loop was intentionally avoided to prevent holding all network streams open until function return

### DIFF
--- a/internal/node/mcp_handlers.go
+++ b/internal/node/mcp_handlers.go
@@ -591,56 +591,76 @@ func (n *SamNode) fetchRemoteToolCatalogue(ctx context.Context, targetPeer peer.
 			connectService = api.MCPServicePrefix + connectService
 		}
 
-		n.preparePeerAddrs(ctx, targetPeer)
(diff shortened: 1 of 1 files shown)


aojea:
@IbrahimAhmed8 check the reviews comments and please squash any change into one commit

IbrahimAhmed8:
> @IbrahimAhmed8 check the reviews comments and please squash any change into one commit

@aojea I have actioned the comments and squashed the changes

aojea:
Thanks 🥇

PR #329, opened 2026-08-31
More CI flake fixes

aojea:
- biscuit evaluation on ci takes longer because the race detector is enabled so override the timeout like we do in other testing
- plain sleeps are wrong, this is tech debt we need to fix by implementing active polling and not relying on waiting async expecting thigs to happen the way we want

### DIFF
--- a/api/datalog_test.go
+++ b/api/datalog_test.go
@@ -333,7 +333,7 @@ func TestControlPlaneStaticTimeCheck(t *testing.T) {
 			}})
 
 			tok, _ := builder.Build()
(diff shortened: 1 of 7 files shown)


aojea:
> Error: ../../../go/pkg/mod/github.com/multiformats/go-multihash@v0.2.3/register/blake3/multihash_blake3.go:15:2: lukechampine.com/blake3@v1.4.1: read "https://proxy.golang.org/lukechampine.com/blake3/@v/v1.4.1.zip": stream error: stream ID 433; INTERNAL_ERROR; received from peer

plenty of internet connectiviy issues today in CI

PR #347, opened 2026-09-01
feat(node): add A2A support

kaisoz:
## What

Adds first-class support for A2A (Agent2Agent protocol) services to sam-node, so agents on the mesh can speak the standard inter-agent protocol with SAM's attested-label sovereignty gating underneath.

- **`a2a` service type**: new `SERVICE_TYPE_A2A` enum value and string mapping. An a2a service is a local agent process registered with `target_url` (command backends are rejected); the node reverse-proxies to it, same shape as inference. Ingress authorization, DHT announce/discovery, and registration work through the existing type-generic paths.
- **Label-gated egress**: the raw egress proxy (`/sam/{peer}/a2a/{svc}/...`) now honors `X-Sam-Required-Labels: key=value[,key=value]`. The caller's node verifies the provider's control-plane-attested labels and refuses fail-closed (HTTP 403) **before any payload leaves the node**; malformed headers get 400. The header is stripped before forwarding. The type match is case-insensitive to mirror ingress parsing.
- **Agent-card rewrite**: responses to `GET .../.well-known/agent-card.json` are rewritten on the caller's node so a stock A2A client works against the mesh unmodified — interface URLs point back at the mesh path, transports the mesh cannot carry (gRPC) are dropped, and streaming is advertised off until verified over libp2p.

All A2A logic is confined to `internal/node/a2a_service.go`; other files carry only dispatch (enum/string cases, one factory case, two insertions in the egress proxy).

## Why

A2A is the emerging standard for inter-agent communication, but it has no notion of data sovereignty. Routing it through SAM gives every A2A call a fail-closed, infrastructure-level guarantee about *where* the counterpart runs (e.g. `region=eu-west-1`), attested by the control plane at enrollment — something neither prompts nor the remote agent's self-description can provide.

## Demo

```
curl http://localhost:8080/sam/<peer>/a2a/<svc>/ \
  -H "X-Sam-Authentication: Bearer <token>" \
  -H "X-Sam-Required-Labels: region=eu-west-1" \
  -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","id":1,"method":"message/send","params":{"message":{...}}}'
```

A provider that cannot prove the label is refused with 403 before the request body leaves the caller's node.

## Testing

- Unit tests for the service factory, the egress hook (gate engagement, malformed labels, invalid peer, case variance, non-a2a passthrough) and the card rewrite (URL rewrite, gRPC drop, streaming off, non-200/content-encoded skip).
- Integration test `TestA2ACUJ`: two nodes, provider enrolled with `--labels region=eu`; asserts the rewritten card, a label-matched `message/send` succeeding across the mesh, a mismatched label refused with 403 with the backend request counter unchanged, and the labels header never reaching the backend.

### DIFF
--- a/development/examples/chat-a2a/agent.py
+++ b/development/examples/chat-a2a/agent.py
@@ -96,7 +96,7 @@ handler = DefaultRequestHandler(
     agent_card=agent_card,
 )
 # Starlette over FastAPI: the SDK generates the routes, so FastAPI would add nothing.
-# JSON-RPC at "/": the mesh card rewrite drops URL subpaths, so clients land on the root.
+# JSON-RPC at "/": the mesh card regeneration drops URL subpaths, so clients land on the root.
 app = Starlette(
     routes=[
         *create_jsonrpc_routes(request_handler=handler, rpc_url="/"),
--- a/development/examples/chat-a2a/agent.py
+++ b/development/examples/chat-a2a/agent.py
@@ -1,0 +1,108 @@
+"""Gemini-backed A2A chat agent hosted by a node in the local dev mesh."""
+import os
+import time
+import uuid
+
+import uvicorn
+from a2a.server.agent_execution.agent_executor import AgentExecutor
+from a2a.server.agent_execution.context import RequestContext
+from a2a.server.events.event_queue import EventQueue
+from a2a.server.request_handlers import DefaultRequestHandler
+from a2a.server.routes import create_agent_card_routes, create_jsonrpc_routes
+from a2a.server.tasks.inmemory_task_store import InMemoryTaskStore
+from a2a.types import (
+    AgentCapabilities,
+    AgentCard,
+    AgentInterface,
+    AgentSkill,
+    Message,
+    Part,
+    Role,
+)
+from google import genai
+from google.genai import types
+from starlette.applications import Starlette
+
+PORT = 7777
+MODEL = os.environ.get("GEMINI_MODEL", "models/gemini-3.6-flash")
+
+class ChatExecutor(AgentExecutor):
+    """One Gemini chat session per A2A contextId; the session carries the history."""
+
+    def __init__(self):
+        self.gemini = genai.Client()
+        self.chats = {}
+
+    async def execute(self, context: RequestContext, event_queue: EventQueue) -> None:
+        chat = self.chats.get(context.context_id)
+        if chat is None:
+            # Gemini 3 Flash defaults to thinking_level=high, which dominates latency.
+            chat = self.gemini.aio.chats.create(
+                model=MODEL,
+                config=types.GenerateContentConfig(
+                    thinking_config=types.ThinkingConfig(thinking_level="minimal")
+                ),
+            )
+            self.chats[context.context_id] = chat
+        started = time.monotonic()
+        reply = await chat.send_message(context.get_user_input())
+        print(
+            f"[chat] context={context.context_id} gemini took "
+            f"{time.monotonic() - started:.1f}s usage={reply.usage_metadata}",
+            flush=True,
+        )
+        await event_queue.enqueue_event(
+            Message(
+                role=Role.ROLE_AGENT,
+                message_id=str(uuid.uuid4()),
+                parts=[Part(text=reply.text or "")],
+                context_id=context.context_id,
+                task_id=context.task_id,
+            )
+        )
+
+    async def cancel(self, context: RequestContext, event_queue: EventQueue) -> None:
+        pass
+
+
+agent_card = AgentCard(
+    name="chat",
+    description="Gemini-backed conversational agent; remembers the conversation per contextId",
+    version="0.1.0",
+    capabilities=AgentCapabilities(streaming=False),
+    default_input_modes=["text"],
+    default_output_modes=["text"],
+    skills=[
+        AgentSkill(
+            id="chat",
+            name="chat",
+            description="Multi-turn small talk",
+            tags=["chat"],
+            examples=["hi, my name is Ada", "what is my name?"],
+        )
+    ],
+    supported_interfaces=[
+        AgentInterface(
+            protocol_binding="JSONRPC",
+            protocol_version="1.0",
+            url=f"http://127.0.0.1:{PORT}/",
+        )
+    ],
+)
+
+handler = DefaultRequestHandler(
+    agent_executor=ChatExecutor(),
+    task_store=InMemoryTaskStore(),
+    agent_card=agent_card,
+)
+# Starlette over FastAPI: the SDK generates the routes, so FastAPI would add nothing.
+# JSON-RPC at "/": the mesh card rewrite drops URL subpaths, so clients land on the root.
+app = Starlette(
+    routes=[
+        *create_jsonrpc_routes(request_handler=handler, rpc_url="/"),
+        *create_agent_card_routes(agent_card=agent_card),
+    ]
+)
+
+if __name__ == "__main__":
+    uvicorn.run(app, host="0.0.0.0", port=PORT)
--- a/development/examples/chat-a2a/agent.py
+++ b/development/examples/chat-a2a/agent.py
@@ -24,7 +24,7 @@ from google.genai import types
 from starlette.applications import Starlette
 
 PORT = 7777
-MODEL = os.environ.get("GEMINI_MODEL", "models/gemini-3.6-flash")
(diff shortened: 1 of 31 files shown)


internal/node/a2a_service.go
```diff
@@ -0,0 +1,159 @@
...
+		caps["streaming"] = false
+	}
+	out, err := json.Marshal(card)
+	if err != nil {
+		return err
+	}
+	resp.Body = io.NopCloser(bytes.NewReader(out))
+	resp.ContentLength = int64(len(out))
+	resp.Header.Set("Content-Length", strconv.Itoa(len(out)))
+	return nil
```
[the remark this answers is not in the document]
kaisoz: `httputil.ReverseProxy` removes `TransferEncoding`, before `ModifyResponse` runs, and therefore the outgoing response's framing is determined by the `Content-Length` header set there. Clearing it would be a no-op.

internal/node/sidecar.go
```diff
@@ -653,6 +653,61 @@ func handleDiscoverService(node *SamNode, w http.ResponseWriter, r *http.Request
...
+	if !ok || mw.gateRequest == nil {
+		return r, true
+	}
+	return mw.gateRequest(node, w, r, route)
+}
+
+// egressModifyResponse dispatches to the middleware of the service type in
+// the rewritten path (/{type}/{svc}/...); types without a hook pass through.
+func egressModifyResponse(resp *http.Response) error {
+	parts := strings.SplitN(strings.TrimPrefix(resp.Request.URL.Path, "/"), "/", 2)
```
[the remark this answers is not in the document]
kaisoz: `egressModifyResponse` is only ever installed as `ReverseProxy.ModifyResponse`, and the proxy guarantees `resp.Request` is the outbound request it just sent, so the nil case can't occur in today's code path.

development/examples/chat-a2a/agent.py
```diff
@@ -0,0 +1,108 @@
...
+    async def execute(self, context: RequestContext, event_queue: EventQueue) -> None:
+        chat = self.chats.get(context.context_id)
+        if chat is None:
+            # Gemini 3 Flash defaults to thinking_level=high, which dominates latency.
+            chat = self.gemini.aio.chats.create(
+                model=MODEL,
+                config=types.GenerateContentConfig(
+                    thinking_config=types.ThinkingConfig(thinking_level="minimal")
+                ),
+            )
```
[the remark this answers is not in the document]
kaisoz: `client.aio.chats.create()` is a synchronous factory in `google-genai`: it builds and returns an AsyncChat without doing any I/O. Besides, the example has been run end to end with multi-turn conversations, which would crash on the first `send_message` if create returned a coroutine.

site/content/docs/user/node-configuration.md
```diff
@@ -69,6 +69,13 @@ When configuring `target_url` for `type: inference` services (e.g. Ollama, vLLM,
 * **OpenAI Facade Access**: Clients connecting via the node's local OpenAI Facade (`http://localhost:8080/v1`) request paths like `/v1/chat/completions`. SAM automatically proxies these to the backend's root URL.
 * **Raw Proxy Access**: If bypassing the Facade and making requests directly via the local egress proxy (`/sam/{peer}/inference/{service}`), the request path must include the explicit `/v1` namespace suffix (e.g. `http://localhost:8080/sam/{peer}/inference/{service}/v1/chat/completions`).
 
+### A2A Service Routing
+
+When configuring `type: a2a` services (Agent2Agent protocol agents):
+* **URL backends only**: register the agent's local HTTP endpoint as `target_url`. `command` backends are rejected.
+* **Raw Proxy Access**: remote peers reach the agent at `http://localhost:8080/sam/{peer}/a2a/{service}/...`. The agent card served at `.../.well-known/agent-card.json` is rewritten in transit so its interface URLs point back at this mesh path; transports the mesh cannot carry (gRPC) are dropped and streaming is advertised off.
```
aojea: rewriting things in transit is scary ...  who is creating that card? is not possible to generate the card at the service registration time with the right url?

internal/node/a2a_service.go
```diff
@@ -0,0 +1,159 @@
...
+	if resp.Header.Get("Content-Encoding") != "" {
+		logger.Warnf("[A2A] agent card response is content-encoded; skipping rewrite")
+		return nil
+	}
+	body, err := io.ReadAll(resp.Body)
+	_ = resp.Body.Close()
+	if err != nil {
+		return err
+	}
+	var card map[string]any
```
aojea: we should use the official typed version https://github.com/a2aproject/a2a-go/blob/03b1f8483cc9cbdd3e95567100d95c5509687eb5/a2a/agent.go#L35 so we detect problems at build time and avoid the map[string]any

internal/node/a2a_service.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// a2aCardBaseURL is the context key carrying the caller-facing mesh base URL
+// of an agent-card fetch, set by a2aEgressGate and consumed by the rewrite.
+type a2aCardBaseURL struct{}
+
+// a2aEgressGate runs the caller-side A2A checks on a raw egress request:
+// the fail-closed labels gate and tagging agent-card fetches for rewrite.
+// On refusal it writes the HTTP error itself and returns ok=false.
+func a2aEgressGate(node *SamNode, w http.ResponseWriter, r *http.Request, route egressRoute) (*http.Request, bool) {
```
aojea, line 62: I do not understand the rewrite here, IIRC this is basically creating a fake handler for `/.well-known/agent-card.json` that builds a json at runtime.
But this can be obtained at registration time, that we know the peer id and we know the url that will be reachable in the mesh, so no need to rewrite, just generate a handler and return this cached response

internal/node/a2a_service_test.go
```diff
@@ -0,0 +1,243 @@
...
+		StatusCode: http.StatusOK,
+		Header:     http.Header{},
+		Body:       io.NopCloser(strings.NewReader(card)),
+		Request:    req,
+	}
+	if err := rewriteA2AAgentCard(resp); err != nil {
+		t.Fatal(err)
+	}
+	body, _ := io.ReadAll(resp.Body)
+	var got map[string]any
```
aojea: avoid dealing with these unstructored types that makes it risky to suffer runtime problems ...

aojea:
Pushed two follow-up commits on top of @kaisoz's work (thanks for the great base — all original commits preserved):

**fb5c349 — typed card regeneration via the official SDK, replacing the in-flight rewrite**
- Adopted `github.com/a2aproject/a2a-go/v2` (types + client/server used in tests; the `a2a` types package is stdlib-only, and the client/server packages needed zero new modules — `uuid`/`x/mod`/`x/sync` were already in the graph).
- The caller's node now *impersonates* the card endpoint: it holds the client request, fetches the card from the agent over the mesh, and serves a card regenerated from typed structs pointing at the local mesh path. This removes the `ReverseProxy.ModifyResponse` rewrite, the context-key plumbing for the base URL, and the content-encoding bail-out (the node negotiates identity encoding on its own fetch).
- Regeneration drops the original JWS `signatures`: they sign the card bytes *including interface URLs*, and the mesh URL is caller-relative (every consumer node mints a different base), so no origin signature can ever verify post-regeneration. Verified stock clients don't verify by default (the Go SDK client has no verification code; python is opt-in). Node re-signing with its own key is a possible follow-up for "must-be-signed" client policies.

**a3d9c4f — stock-SDK coverage at every level, and three real bugs it surfaced**
- Integration CUJ now runs the official Go SDK on both ends (`a2asrv` echo agent, `a2aclient` + card resolver through the mesh, ~2s), and a new e2e bats CUJ (`tests/e2e/a2a_mesh.bats`) drives the containerized mesh with the stock **python** `a2a-sdk` on both ends, registering via node config and exercising the labels gate over the real attestation chain.
- Testing with real SDK parsers instead of hand-rolled JSON found:
  1. regenerated cards kept nil required list fields → `"skills":null`, which pydantic-based clients reject for the whole card; now normalized to empty arrays.
  2. the official Go resolver treats a pathful base URL as the card URL itself (python appends `/.well-known/agent-card.json`), so the regenerated card is now also served at the bare service root (safe: JSON-RPC is POST-only).
  3. the Helm bootstrap job had no way to grant `allowed_labels`, so no chart-deployed mesh could enroll a labelled node at all; added `bootstrap.nodeLabels` (default `[]`, fail closed), pinned by chart tests.

All gates green locally: `make lint`, chart tests 31/31, full `internal/node`, integration, and the new e2e CUJ.

PR #350, opened 2026-09-01
enforce freshness boundary on token refresh

### LINKED ISSUE #349: Token refresh challenge is vulnerable to replay attacks
I was looking at the `/refresh` endpoint and noticed there's no freshness check on `req.Timestamp` before we verify the signature in `HandleRefresh`

Right now it just does this:
```go
challengeData := []byte(fmt.Sprintf("%d", req.Timestamp))
ok, err := pubKey.Verify(challengeData, req.ChallengeSignature)
```
Because we don't check how old the timestamp is, anyone who intercepts a valid refresh request can just replay it indefinitely (until the OIDCSessionTTL expires) to keep minting fresh biscuits without needing the private key again. I tested this locally by sending a request with a timestamp from a few hours ago and it still returned a 200 OK with a new token

We should enforce a strict window here and return a 401 if it's too old. I'll raise a PR shortly with the fix and some tests to sort this out

IbrahimAhmed8:
Resolves #349

I used a 5m window to match the existing FreshnessThreshold for mesh events

### DIFF
--- a/api/sam.pb.go
+++ b/api/sam.pb.go
@@ -1674,11 +1674,13 @@ func (x *KeysResponse) GetPublicKeys() [][]byte {
 }
 
 type TokenRefreshRequest struct {
-	state              protoimpl.MessageState `protogen:"open.v1"`
-	ChallengeSignature []byte                 `protobuf:"bytes,1,opt,name=challenge_signature,json=challengeSignature,proto3" json:"challenge_signature,omitempty"`
-	Timestamp          int64                  `protobuf:"varint,2,opt,name=timestamp,proto3" json:"timestamp,omitempty"`
-	unknownFields      protoimpl.UnknownFields
-	sizeCache          protoimpl.SizeCache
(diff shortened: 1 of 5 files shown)


aojea:
Thanks @IbrahimAhmed8, good catch and a clean fix — verified the replay locally and the 5m window matches the mesh `FreshnessThreshold`.

Pushed one follow-up commit (aaaf4a9) on top of yours:
- **Single-use rotation**: every issuance path already persists the latest biscuit, so `/refresh` now refuses any other one. A captured request replayed *inside* the freshness window mints nothing; nodes that lose a rotated token recover via the existing re-enrollment fallback.
- Documented the `TokenRefreshRequest` contract in `sam.proto` (timestamp = Unix milliseconds, signature over its decimal string) and regenerated.
- Folded the freshness cases into a table test (window edges, seconds-vs-millis, zero/negative) plus a byte-identical replay scenario.

All green locally: `make lint`, full controlplane/node/router/integration suites.

PR #359, opened 2026-09-05
resolve goroutine leak in serveOnPipe

### LINKED ISSUE #358: Goroutine leak in serveOnPipe blocks http.Server.Serve on Accept indefinitely
I was looking at the new tunnels from #346 and found a goroutine leak in `internal/sambox/mesh.go`. The problem is with how `singleConnListener` handles disconnects. When the underlying net.Pipe drops, that closure never triggers the listener's Close() method. This means http.Server.Serve just hangs forever on Accept waiting for <-l.closed, which basically leaks a goroutine on every single disconnect.

Here:

```go


func (l *singleConnListener) Accept() (net.Conn, error) {
	var conn net.Conn
	l.accept.Do(func() { conn = l.conn })
	if conn != nil {
		return conn, nil
	}
	<-l.closed
	return nil, net.ErrClosed
}
```

IbrahimAhmed8:
Resolves #358

### DIFF
--- a/internal/sambox/mesh.go
+++ b/internal/sambox/mesh.go
@@ -82,6 +82,7 @@ func (d *AgentDialer) dialMeshService(ctx context.Context, route Route) (net.Con
 // connection.
 func serveOnPipe(h http.Handler) net.Conn {
 	agentSide, boundarySide := net.Pipe()
+	ln := newSingleConnListener(boundarySide)
 	server := &http.Server{
 		Handler: h,
 		// Mirrors the sidecar: bound header reads, but let bodies and responses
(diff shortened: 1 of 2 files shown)


.github/workflows/test_flutter.yaml
```diff
@@ -63,24 +58,23 @@ jobs:
...
       with:
         go-version: ${{ env.GO_VERSION }}
+        cache: false
     - uses: subosito/flutter-action@1a449444c387b1966244ae4d4f8c696479add0b2 # v2
       with:
         channel: 'stable'
-        cache: true
     - name: Set up JDK
-      uses: actions/setup-java@dd06d9cba3e5552c54d9f8ea23572deb30010f7c # v6.0.0
+      uses: actions/setup-java@b6effb05e454b25005698d916606bdc6ffcbf961 # v5.7.0
```
aojea: this change seems unrelated, no?

IbrahimAhmed8: i'll revert that as it was included by mistake during local testing,
apologize

aojea:
@IbrahimAhmed8 can you please rebase?

PR #377, opened 2026-09-09
node: make the backend probe timeout configurable

### LINKED ISSUE #376: sam-node: command-spawned backend health-check deadline (~2.4s) has no configurable timeout, and is tighter than realistic MCP server startup costs
**What's missing**

The `command`-spawned backend health check in `sam-node` (used when a `type: mcp` service in `sam-node.yaml` spawns a local subprocess) has a deadline that isn't exposed as a configurable flag anywhere I could find (checked `sam-node run --help` and `sam-node.yaml`'s documented schema).

**Why this matters in practice**

Measured the actual window: from node start to the registry giving up (`backend did not answer: context deadline exceeded`) is **~2.4s** (timestamps from `node-producer.log`, node start to give-up).

That's tighter than the cold-import cost of moderately-featured MCP server stacks, measured directly on this machine:
- `import fastmcp` alone: **~2.7s**
- `import agentmesh.governance` (Microsoft's Agent Governance Toolkit) alone: **~5.4s**
- `npx -y @modelcontextprotocol/server-everything` (SAM's *own* reference/test server, referenced in `sam-node skill show` as `mcp://everything/get-sum`) cold-starting to answer `initialize`, even warm/cached: **~4.0s**

That last one is the concerning data point: SAM's own documented reference example can't reliably meet its own health-check deadline on Windows. This isn't a workload-specific edge case - it looks like the deadline is just too tight for realistic subprocess startup costs (interpreter + framework import + first handshake), at least on Windows.

**What I'd like**

A `--backend-health-timeout` (or equivalent `sam-node.yaml` field) to raise this deadline for slower-starting backends, so services aren't silently `Registered ... but not advertising` / withheld from the DHT forever with no retry (observed no retry after waiting 20s past the initial failure).

**Workaround used in the meantime**

A minimal stdlib-only stdio<->HTTP relay that starts near-instantly and proxies to an already-warm HTTP-mode MCP server, so the slow import cost happens off the health-check's critical path. Works, but shouldn't be necessary for a straightforward Python (or Node) MCP server.

Related to the retry/session issues in #375, but distinct - this is specifically about the health-check deadline itself being too tight, independent of whatever's causing that one.

Happy to share the timing methodology / measurement scripts if useful.

fer-marino:
Fixes #376.

sam-node.yaml's command-spawned MCP services are given a hard-coded 2s (dhtProbeTimeout) to answer an MCP initialize before the service is registered but withheld from advertisement ("backend did not answer: context deadline exceeded"), with no retry observed afterwards.

2s is tighter than the cold-start cost of realistic backends. Measured directly: a bare `import fastmcp` (Python) takes ~2.7s, and even SAM's own reference example, `npx -y @modelcontextprotocol/server-everything`, takes ~4.0s to answer initialize on Windows even warm/cached - SAM's own documented reference server cannot reliably meet its own default.

Adds a new ServiceRegistry.SetBackendProbeTimeout, wired from a new --backend-probe-timeout flag (0 keeps the existing 2s default, matching the convention already used by --dht-max-record-age and similar flags). Backward compatible: NewServiceRegistry's default is unchanged, and every existing call site that doesn't pass the new flag behaves exactly as before (verified: all pre-existing internal/node tests pass unmodified, including 6 that fail identically on unmodified main - confirmed via git stash - so unrelated to this change).

Also fixes three of the four node.Options{} construction sites in cmd/sam-node/main.go that already wire NewServiceRegistry-adjacent flags (DHTMaxRecordAge et al.) but were missing this one; the fourth (join-only enrollment path) doesn't call RegisterStaticServices and is out of scope.

Adds TestServiceRegistry_BackendProbeTimeoutIsConfigurable, using a new slowProbingService test fake that (unlike the existing probingService) actually respects context deadlines, so it can demonstrate: the same slow backend fails to advertise under the default timeout and succeeds once given more time.

### DIFF
--- a/internal/node/options.go
+++ b/internal/node/options.go
@@ -142,6 +142,9 @@ func (o *Options) Default() {
 	if o.NodeConfig == nil {
 		o.NodeConfig = &NodeConfigComplete{}
 	}
+	if o.BackendProbeTimeout <= 0 {
+		o.BackendProbeTimeout = defaultDHTProbeTimeout
+	}
 }
 
 // Validate verifies that the required options are provided and valid.
--- a/internal/node/options.go
+++ b/internal/node/options.go
@@ -78,6 +78,14 @@ type Options struct {
 	PolicySyncInterval time.Duration
 	// PolicySyncJitter specifies the maximum jitter delay when scheduling policy syncs on event broadcasts.
 	PolicySyncJitter time.Duration
+	// BackendProbeTimeout bounds how long a command-spawned service backend
+	// (sam-node.yaml's `command`, spawned as a local subprocess) is given to
+	// answer before the service is registered but withheld from
+	// advertisement. Zero uses the library default (2s). Raise this for
+	// backends with slower cold-start/import costs than that - the default
+	// is tight enough that even simple interpreted-language MCP servers can
+	// miss it on first spawn.
+	BackendProbeTimeout time.Duration
 }
 
 // Default applies default values to Options if they are not specified.
--- a/internal/node/service_registry.go
+++ b/internal/node/service_registry.go
@@ -38,7 +38,7 @@ type backendProber interface {
 	Probe(ctx context.Context) error
 }
 
-// dhtProbeTimeout is the default bound on one backend probe before
+// defaultDHTProbeTimeout is the default bound on one backend probe before
 // advertising, used when a ServiceRegistry isn't given an explicit
 // BackendProbeTimeout (see Options.BackendProbeTimeout / --backend-probe-
 // timeout). Command-spawned backends (sam-node.yaml's `command`, launched as
@@ -46,7 +46,7 @@ type backendProber interface {
 // request - a moderately-featured interpreted-language MCP server's own
 // import/startup cost alone can exceed 2s - so this is a floor, not
 // something every backend is expected to meet.
-const dhtProbeTimeout = 2 * time.Second
+const defaultDHTProbeTimeout = 2 * time.Second
 
 // advertisable reports whether a service is fit to be published to the DHT.
 //
@@ -78,7 +78,7 @@ type ServiceRegistry struct {
 	reprovideNow func()
 
 	// backendProbeTimeout bounds each backend probe in advertisable. Defaults
-	// to dhtProbeTimeout; override with SetBackendProbeTimeout.
+	// to defaultDHTProbeTimeout; override with SetBackendProbeTimeout.
 	backendProbeTimeout time.Duration
 }
 
@@ -86,12 +86,12 @@ func NewServiceRegistry(d dhtProvider) *ServiceRegistry {
 	return &ServiceRegistry{
 		services:            map[string]Service{},
 		dht:                 d,
-		backendProbeTimeout: dhtProbeTimeout,
+		backendProbeTimeout: defaultDHTProbeTimeout,
 	}
 }
 
 // SetBackendProbeTimeout overrides the default backend probe timeout
-// (dhtProbeTimeout). A zero or negative duration is a no-op, so callers can
+// (defaultDHTProbeTimeout). A zero or negative duration is a no-op, so callers can
 // pass an unset Options.BackendProbeTimeout straight through without an
 // explicit zero-check.
 func (r *ServiceRegistry) SetBackendProbeTimeout(d time.Duration) {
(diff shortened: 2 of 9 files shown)


internal/node/service_registry.go
```diff
@@ -38,7 +38,14 @@ type backendProber interface {
...
-// dhtProbeTimeout bounds one backend probe before advertising.
+// dhtProbeTimeout is the default bound on one backend probe before
+// advertising, used when a ServiceRegistry isn't given an explicit
+// BackendProbeTimeout (see Options.BackendProbeTimeout / --backend-probe-
+// timeout). Command-spawned backends (sam-node.yaml's `command`, launched as
+// a local subprocess) can need longer than this to answer their first
+// request - a moderately-featured interpreted-language MCP server's own
+// import/startup cost alone can exceed 2s - so this is a floor, not
+// something every backend is expected to meet.
 const dhtProbeTimeout = 2 * time.Second
```
aojea: let's make the variable most meaningful `defaultDHTProbeTimeout`

aojea:
sing the CLA and address the review commentds and we are good to go, thanks

internal/node/service_registry.go
fer-marino: Done in 81db9e9 - renamed to `defaultDHTProbeTimeout`. Also addressed the other two review comments in the same commit (fall back to it for a zero-initialized `ServiceRegistry`, and fixed a timer leak in the test).

internal/node/service_registry_test.go
```diff
@@ -274,3 +275,82 @@ func TestServiceRegistry_ReprovideResumesWhenBackendRecovers(t *testing.T) {
...
+}
+
+func (p *slowProbingService) Probe(ctx context.Context) error {
+	select {
+	case <-time.After(p.delay):
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
```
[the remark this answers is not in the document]
fer-marino: Good catch, fixed in 81db9e9 exactly as suggested - time.NewTimer + defer timer.Stop().

internal/node/service_registry.go
```diff
@@ -69,15 +76,40 @@ type ServiceRegistry struct {
...
+	defer r.mu.Unlock()
+	r.backendProbeTimeout = d
+}
+
+// probeTimeout returns the current backend probe timeout.
+func (r *ServiceRegistry) probeTimeout() time.Duration {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	return r.backendProbeTimeout
+}
```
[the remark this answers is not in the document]
fer-marino: Good catch, fixed in 81db9e9 - probeTimeout() now falls back to dhtProbeTimeout (renamed defaultDHTProbeTimeout per aojea's comment below) when backendProbeTimeout is <= 0.

internal/node/service_registry.go
```diff
@@ -69,13 +76,45 @@ type ServiceRegistry struct {
...
+	defer r.mu.Unlock()
+	r.backendProbeTimeout = d
+}
+
+// probeTimeout returns the current backend probe timeout, falling back to
+// defaultDHTProbeTimeout for a zero-initialized registry (e.g. a struct
+// literal built directly in a test, bypassing NewServiceRegistry) so it
+// behaves the same as a properly constructed one rather than timing out
+// every probe immediately.
+func (r *ServiceRegistry) probeTimeout() time.Duration {
```
aojea: we don't use setters and getters, see internal/node/options.go to see how the config is populated

fer-marino: Good point, done in f0a1401 - dropped both SetBackendProbeTimeout and probeTimeout(). NewServiceRegistry now takes backendProbeTimeout directly and defaults it once at construction if <= 0, matching how options.go populates config; the field is never mutated afterwards so reads need no lock either.

aojea:
I prefer to squash the commits, having the intermediate steps merged are not useful because they are discarded in commits later

fer-marino:
Sure. That is usually my preference too

internal/node/options.go
aojea, line 92: default values are set here

fer-marino: Done in deadf0e - BackendProbeTimeout now defaults here too, alongside every other Options field. NewServiceRegistry keeps its own <=0 fallback as well, since it's also constructed directly in tests that bypass Options entirely, but by the time it's called from Start() the value has already been normalized here.

aojea:
/gemini review

one last pass just to be sure, but this LGTM

aojea:
@fer-marino since you have to rebase, what about considering the suggestion in https://github.com/google/sam/pull/377#discussion_r3976254872 to avoid flakes? I think is reasonable

fer-marino:
Both done, pushed as 1d047d1:
- Rebased onto main (had a couple of real conflicts against #382's labels-config move, resolved in the two "fix rebase conflict"/follow-up commits).
- Applied your suggested deterministic rewrite of the probe-timeout test verbatim - slowProbingService is now deadline-inspecting/context-controlled instead of a real timer, no more real-time sleeps.

aojea:
Thanks

PR #379, opened 2026-09-09
fix: mcp_service - don't drop the last response on backend EOF (#375)

### LINKED ISSUE #375: sam-node: tools/call succeeds on the backend but consumer sees EOF (producer->consumer relay drops the response)
**Environment**
- SAM `v0.1.0-alpha.9` (Windows x86_64 binaries)
- Windows 11, local `sam-one` standalone mesh (control plane + router, sqlite backend)
- Two `sam-node` instances joined via bootstrap token, `--allow-loopback`
- Producer node's `sam-node.yaml` registers a `type: mcp` service via `command` (stdio subprocess)

**What's broken**

A `tools/call` (and `tools/list`) issued from a consumer node via `call_remote_tool` / `find_remote_tools` against a service on a different node consistently fails with:

```
failed after 3 retries: failed to call tool mcp://<service>/<tool>: calling "tools/call": EOF
```

This reproduces with two independent clients (the bundled `mcp-client.exe`, and Python's `langchain-mcp-adapters` talking to the consumer node's `/mcp` endpoint) - not a client bug.

**Root-caused to the producer→consumer response relay, not the backend**

Instrumented the `command`-spawned subprocess (a stdio<->HTTP relay to a warm backend) to log every raw line in/out with timestamps. Evidence:

1. The subprocess produces the *correct* `tools/call` response, twice (once per retry), matched to the request:
   ```
   IN  {"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"name":"...","arguments":{...}}}
   OUT ok {"jsonrpc":"2.0","id":3,"result":{"content":[{"text":"...correct payload..."}],"isError":false,...}}
   ```
2. The producer's own log shows a clean write-back to the consumer at the exact moment of that success, no error:
   ```
   2026-09-09T09:17:09.770+0200  Stream Accounting  {"target": "mcp://<service>", "bytes_read": 1218, "bytes_written": 834}
   ```
3. The consumer logs EOF **1ms later**, for the same call:
   ```
   2026-09-09T09:17:09.771+0200  [MCP] Tool call failed, retrying in 2s: ... calling "tools/call": EOF
   ```

Repeats identically on the retry 2s later. The producer believes it wrote a complete response; the consumer sees EOF at essentially the same instant. The backend, the subprocess relay, and the client are demonstrated correct by direct log evidence - the gap is specifically in the producer→consumer P2P response relay (or the consumer's read of it).

**Positive control (rules out "your backend is unusual")**

Tried the same pattern fronting `@modelcontextprotocol/server-everything` (the reference server `sam-node skill show` itself references, `mcp://everything/get-sum`) instead of our own backend:
- Direct `command: [npx, -y, "@modelcontextprotocol/server-everything"]` (no relay): failed registration outright - `npx` takes ~4s to start even warm, exceeding the backend health-check deadline. (Separately worth noting: this deadline appears to have no configurable flag, and is tight enough that even SAM's own reference example can't reliably meet it on Windows.)
- Via the same relay pattern, fronting a warm `server-everything` instance: different failure signature during health-check - `unmarshaling jsonrpc message: EOF` on the first attempt, then `HTTP Error 400: Bad Request` on retries.

Two different backends, two different but both-broken symptoms around retries/session handling for `command`-spawned services.

**Ask**

Would appreciate a pointer to whether this is a known issue (didn't find an obvious match in the open issues) and, if useful, I'm happy to share the full wire-level debug log and minimal repro scripts (a ~50-line stdlib-only stdio<->HTTP relay + sam-node.yaml).

Tested on Windows only - haven't yet confirmed whether this reproduces on Linux.

fer-marino:
## Summary

Fixes #375.

`HandleStreamPassThrough` proxies JSON-RPC messages between a client-facing P2P stream and a backend MCP connection. It used to shut the whole thing down - `s.Close()` - the instant *either* leg reported *any* error, including the completely normal case of a one-shot HTTP-style backend answering one request and then closing its side (a clean `EOF`, not a failure).

`network.Stream.Close`'s own documented contract:
> Close may be asynchronous and *does not* guarantee receipt of the data... If acknowledgment is required, the caller must call `CloseWrite`, then wait on the stream for a response (or an EOF), then call `Close()`.

The old code skipped exactly that: the backend-read goroutine would `Write` the response to the client, loop back, get `EOF` from the backend, and immediately trigger the deferred `s.Close()` a few instructions later - closing right behind a write that had already reported success. On the QUIC transport, `Close()` additionally cancels the read side over the wire (`CancelRead` sends a `STOP_SENDING` frame to the peer), which a plain TCP+yamux stream's `CloseRead` explicitly does not do (its doc says "Remote is not notified"). That's consistent with the original report: the producer's write reports success, but the response never fully reaches the consumer, which sees `EOF` instead.

## Fix

The backend leg ending now only half-closes *our* write side to the client (`s.CloseWrite()`) and stops the backend->client relay goroutine; it no longer touches the read side or frees the stream. The client leg - the client itself finishing its read and hanging up, or a genuine transport error - is what triggers the final `s.Close()`. A 5s drain timeout (`passThroughDrainTimeout`) bounds the wait so a client that never hangs up can't leak the stream indefinitely.

## Test plan

- Added `TestHandleStreamPassThrough_BackendEOFDoesNotDropInFlightResponse` in `gate_test.go`, which drives 50 real `CallTool` round trips over a **real local QUIC connection** between two `SamNode`s (a new `startBareQUICNode` helper - `startBareNode`'s existing TCP+yamux transport can't exercise this: yamux's `CloseRead` is local-only per its own doc comment, so it structurally can't reproduce the wire-level race). It passes reliably with the fix.
- Being transparent about the test's limits: run against the pre-fix code, it did **not** reproduce the race on loopback across 1500+ iterations (with and without `GOMAXPROCS=1`), so it is not proof this fixes the exact wire-level interaction seen in the original report - only that the new code follows `network.Stream`'s documented safe-shutdown sequence and behaves correctly across many round trips. Localhost loopback has effectively no latency/jitter, which likely closes the race window that a real multi-host deployment (or a relayed connection) would have. `go build ./...`, `go vet ./...`, and the full `internal/node` suite pass (the 6 pre-existing unrelated failures in that package are present identically on unmodified `main`).
- [ ] Maintainer confirmation against the original multi-host/production repro would be the strongest signal this actually closes out #375.

### DIFF
--- a/internal/node/mcp_service.go
+++ b/internal/node/mcp_service.go
@@ -203,7 +203,13 @@ func (m *MCPService) HandleStreamPassThrough(s network.Stream) {
 	// the final s.Close() at the top of this function. passThroughDrainTimeout
 	// bounds that wait so a client that never hangs up can't leak the stream
 	// forever.
-	clientErrc := make(chan error, 1)
+	//
+	// clientErrc is buffered for 2, not 1: a client write failure below is
+	// also a client-leg error (the stream to the client is dead, so there is
+	// nothing left to drain for), and with both goroutines able to send, an
+	// unlucky interleaving where the main select has already consumed one
+	// value could otherwise leave the second sender blocked forever.
+	clientErrc := make(chan error, 2)
 
 	go func() {
 		for {
@@ -217,6 +223,7 @@ func (m *MCPService) HandleStreamPassThrough(s network.Stream) {
 			}
 			if err := clientConn.Write(ctx, msg); err != nil {
 				logger.Debugf("[MCPService] %s: client write error: %v", m.info.Name, err)
+				clientErrc <- err
 				return
 			}
 		}
--- a/internal/node/mcp_service.go
+++ b/internal/node/mcp_service.go
@@ -140,6 +140,13 @@ var preflightMethodsUnsupportedByPassThrough = map[string]bool{
 	"server/discover": true,
 }
 
+// passThroughDrainTimeout bounds how long HandleStreamPassThrough keeps a
+// client-facing stream open after the backend leg has ended, waiting for the
+// client to finish reading the last response and hang up on its own. See the
+// comment on the drain logic in HandleStreamPassThrough for why this wait
+// exists at all.
+const passThroughDrainTimeout = 5 * time.Second
+
 // HandleStreamPassThrough connects to the backend and proxies JSON-RPC messages.
 func (m *MCPService) HandleStreamPassThrough(s network.Stream) {
 	defer func() {
@@ -148,23 +155,14 @@ func (m *MCPService) HandleStreamPassThrough(s network.Stream) {
 		}
 	}()
 
-	var backendTransport mcp.Transport
-	var closeTransport func()
-
 	backendTransport, err := m.backendTransport()
 	if err != nil {
 		logger.Errorf("[MCPService] %s: %v", m.info.Name, err)
 		return
 	}
-	closeTransport = func() {} // fresh per stream for URL; shared bridge is never closed
 
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
-	defer func() {
-		if closeTransport != nil {
-			closeTransport()
-		}
-	}()
 
 	backendConn, err := backendTransport.Connect(ctx)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/node/mcp_service.go
```diff
@@ -180,49 +178,85 @@ func (m *MCPService) HandleStreamPassThrough(s network.Stream) {
...
-				errc <- err
-				return
-			}
-		}
-	}()
-
-	<-errc
+	select {
+	case <-clientErrc:
+	case <-time.After(passThroughDrainTimeout):
```
aojea, line 277: This has to start counting once one of the sessions has finish, right now it starts when the session start, you need to signal when each session finish and then you wait for it to drain inside 
```go
backendDone := make(chan struct{})
go func() {
    defer close(backendDone)
    // backend -> client relay, CloseWrite on read error
}()
...
select {
case <-clientErrc:
case <-backendDone:
    select {
    case <-clientErrc: // wait for the other side to drain
    case <-time.After(passThroughDrainTimeout):
        logger.Debugf(...)
    }
}
```

aojea: You also need to add another test that covers that scenario, a session that takes longer than `passThroughDrainTimeout` will fail with current PR

fer-marino: Both fixed in 57b5c53 - added the backendDone channel exactly as you sketched (a session that finishes on the client side first returns immediately, the drain wait only starts once the backend leg ends), and added TestHandleStreamPassThrough_SlowBackendDoesNotHitDrainTimeout for the scenario you flagged. Confirmed it actually catches the bug: reverted just the backendDone gating locally and the new test fails with the exact symptom (CallTool: calling "tools/call": EOF) before re-applying the fix.

PR #380, opened 2026-09-09
docs: node-configuration - target_url works for type: mcp, not just H…

fer-marino:
…TTP/inference/A2A

The property table and example config only showed command (stdio subprocess) for MCP services, and described target_url as HTTP/inference/A2A only. That's not what the code does: buildRegisterRequest maps target_url to a backend regardless of service type, and MCPService.backendTransport() proxies to it the same way for MCP as any other type - verified end to end against a real Streamable HTTP MCP server, not just by reading the source.

Adds a second MCP example (HTTP backend via target_url, no subprocess) and clarifies the two backend kinds are mutually exclusive.

### DIFF
--- a/site/content/docs/user/node-configuration.md
+++ b/site/content/docs/user/node-configuration.md
@@ -25,11 +25,17 @@ version: "v1alpha1"
 
 # 1. Define Local Services
 services:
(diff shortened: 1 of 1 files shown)


aojea:
Thanks

PR #381, opened 2026-09-09
mobile: support attested labels on the Android node

kaisoz:
The mobile FFI had no way to declare labels, so a phone always enrolled unlabeled and any call against it with `required_labels` was refused by the caller's labels gate. Getting the label flow to work end to end on the emulator also surfaced several bugs in the app and gaps in the kind dev setup, fixed here.

**Labels**
- `api.ParseLabels`: the `key=value` parser moves from `cmd/sam-node` into `api/labels.go` so the CLI and the FFI share it; the CLI calls it directly.
- `EnrollNode` and `MobileConfig` gain `labels`. Enrollment mints them into the Biscuit; the FFI keeps a copy in the app's data directory and reuses it on start and renewal, since the app only shows the labels field on its enrollment screen.
- The enrollment screen gains an optional labels field.

**App fixes**
- Every enroll path failed with "Illegal argument in isolate message": `Isolate.run` closures inside State methods captured `this` and its `DynamicLibrary`. They now run from top-level functions.
- A second Start tap tore down the embedded MCP server; `start()` is guarded and errors propagate.
- External MCP fields use hints instead of prefilled values.

**Kind dev setup**
- Dex lists the app's device-flow and loopback callbacks; `bootstrap.nodeLabels={*}` so dev nodes can attest labels; debug builds allow cleartext http to the kind cluster.
- The tmux log window shows the console, control plane and Dex URLs.

**Docs**: the mobile README and site page no longer claim the `mobile-ffi-*` targets copy into `jniLibs`, give the right x86_64 paths, and describe how the app carries labels across starts.

Tests: `api.ParseLabels` unit test; the FFI lifecycle test asserts the enroll request carries the labels, that they are persisted, and that a label-less `StartNode` falls back to them; a negative test checks `StartNode` rejects malformed labels.

### DIFF
--- a/api/labels.go
+++ b/api/labels.go
@@ -87,6 +87,32 @@ func ValidateLabels(labels map[string]string) error {
 	return nil
 }
 
+// ParseLabels parses a comma-separated "key=value" list (the wire/CLI form of
+// a label set) into a label map; an empty string means no claims. Parsing is
+// syntax only — run the result through ValidateLabels.
+func ParseLabels(s string) (map[string]string, error) {
+	if s == "" {
+		return nil, nil
+	}
+	labels := make(map[string]string)
+	for _, part := range strings.Split(s, ",") {
+		part = strings.TrimSpace(part)
+		if part == "" {
+			continue
+		}
+		k, v, ok := strings.Cut(part, "=")
+		if !ok {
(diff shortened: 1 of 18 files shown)


kaisoz:
This needs to be reworked as we decided to move the labels to the config from the CLI flags

PR #386, opened 2026-09-10
prevent concurrent close panics in channelListener

IbrahimAhmed8:
The channelListener.Close() method in internal/sambox/gateway.go contains a race condition. If multiple goroutines attempt to close the listener concurrently during disconnections a panic occurs for closing a closed channel, which crashes the node

This patch replaces the unsafe select block with sync.Once to guarantee the channel closes exactly once and avoids the panic completely


aojea:
> This patch replaces the unsafe select block with sync.Once to guarantee the channel closes exactly once and avoids the panic completely

This is only called from one place today so panic is not possible, but better safe than sorry

PR #387, opened 2026-09-11
fix: command-backed MCP services get one subprocess per session, not a shared multiplexed bridge

fer-marino:
**Reworked following @aojea's review below** - the original approach (an unbounded per-subscriber queue on a shared `StdioBridge`) has been replaced entirely. History of how this PR got here is preserved in the commits and the comment thread; this description now describes the current state.

The bug is in the sharing model, not the buffer. `StdioBridge` multiplexed every session of a command backend over one child process and one stdout stream, broadcasting every line to every subscriber and relying on each session's own JSON-RPC id to sort out which reply was whose. That doesn't hold: the go-sdk client numbers requests from 1 per connection, so two concurrent sessions on the same bridge could both send `id: 1` and each read the other's reply - a silent cross-session data leak, not just the drop bug this PR originally set out to fix. The queue fix kept all of that and added its own failure mode: a peer that stops reading its stream retains every other session's traffic indefinitely (unbounded growth instead of a drop).

Stdio MCP is single-session by spec, and no SDK multiplexes it, because a server-initiated request (sampling, elicitation) has no attributable destination on a shared process. So instead of teaching the bridge a per-session id space, `MCPService.backendTransport` now builds a fresh `mcp.CommandTransport` - one subprocess per call - for command backends, the same "fresh transport per session" shape it already used for URL backends via `StreamableClientTransport`. `Probe`, `Tools` and `HandleStreamPassThrough` are unchanged: they only ever see an `mcp.Transport`.

`StdioBridge` and `bridgeTransport` are deleted. Command backends lose the local SSE/POST HTTP ingress route `StdioBridge.ServeHTTP` used to serve (`baseService.Init` no longer builds a handler for them) - the mesh stream path is what actually matters for cross-partner discovery/invocation and is unaffected. `Handler()`'s existing callers already treat a nil handler as "no local ingress for this service" (a 404), not a crash. Restoring a local-ingress equivalent is left for a follow-up, per @aojea's own scoping below.

New tests (`mcp_service_command_test.go`) pin the property this exists for: two concurrent command-backend sessions, both using `id: 1` the way independent go-sdk clients naturally would, each only ever see their own subprocess's replies.

Tests: `go test ./internal/node/...` - full suite unaffected beyond the five pre-existing, unrelated Windows-environment failures noted in the original description (reproduce identically on unmodified main). The new tests pass reliably under repeated runs (`-count=10`).


internal/node/stdio_bridge.go
```diff
@@ -200,27 +210,99 @@ func (b *StdioBridge) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+// queue has been closed and fully drained.
+func (q *subscriberQueue) pop(ctx context.Context) (line string, ok bool, err error) {
+	for {
+		q.mu.Lock()
+		if len(q.buf) > 0 {
+			line = q.buf[0]
+			q.buf = q.buf[1:]
+			q.mu.Unlock()
+			return line, true, nil
+		}
```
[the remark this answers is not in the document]
fer-marino: Good catch - fixed in d44580d, exactly as suggested (clear the popped slot, nil out buf on full drain). Added a regression test that confirms it: a second reference into the backing array reads back the zero value after its line is popped, and buf is nil after a full drain - checked it actually fails without the fix (backing array slot still held the popped string).

internal/node/stdio_bridge.go
```diff
@@ -200,27 +210,110 @@ func (b *StdioBridge) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+// push is called from the bridge's single stdout-scanning goroutine (with
+// StdioBridge.mu held) and must never block or drop a line. pop is called by
+// the session's own reader; it blocks until a line is queued, the bridge
+// closes the queue, or ctx is done.
+type subscriberQueue struct {
+	mu     sync.Mutex
+	buf    []string
+	closed bool
+	notify chan struct{} // capacity 1: a "there may be new work" flag, coalesced
+}
```
aojea, line 257: We have a lot of experience in kubernetes with queues and they are really hard to get right, we still have subtle bugs after 10 years so unless is strictly necessary I want to explore other options, AI tools tend. to looks always for complex solutions but if we add more context there usually are simpler ones .... can we have an integration or e2e test as reproducer first, we need to exercise the end to end path to ensure we are covering all edge cases

fer-marino:
Thanks for the review, @aojea — and for the earlier patience on #375 while I chased that one down to its actual cause.

Pushed one more commit (d951715) adding two integration tests that drive a real subprocess through the actual StdioBridge → bridgeTransport path (not the in-memory pipe the other tests use), since that's the level the original bug and the memory-retention fix both actually live at:

- `TestIntegration_ProbeAndLiveSessionConcurrently`: a live session bursting calls without reading between them, racing a continuously-reconnecting probe — the exact scenario Subscribe's doc comment describes. Fails deterministically against the pre-fix code, passes here.
- `TestIntegration_StuckSubscriberDoesNotStallOthers`: one subscriber that connects and never reads, asserting a second healthy subscriber on the same backend is unaffected. I added this one specifically to check the obvious simpler alternative to an unbounded per-subscriber queue — a fixed-size channel per subscriber with a blocking send instead of drop. It's much less code, but this test catches why it doesn't work: the scan loop dispatches to every subscriber under one lock that Send also takes, so one stuck consumer deadlocks the whole bridge for every other session on that backend. Confirmed that failure mode directly (goroutine dump showed the scan loop blocked on the full channel, holding the lock, with a second session's Write() stuck waiting on the same lock). The unbounded queue avoids that by construction — it never blocks the producer, only grows for the one subscriber that isn't keeping up.

This is ready for another look whenever you have a chance.

aojea:
@fer-marino I was looking into this and I think we are working in a direction that will cost us later.

The bug here is in the sharing model, not in the buffer. Subscribe() fans out every stdout line to every session, so a session's channel fills with replies that belong to other sessions. Worse, the go-sdk numbers requests from 1 per connection, so two concurrent sessions on the same bridge both send id: 1 and both receive both replies — session A can consume session B's tools/call result today. An unbounded per-subscriber queue keeps all of that, and adds a new failure mode: a peer that stops reading its stream (HandleStreamPassThrough has no timeout while the backend leg is alive) retains every other session's traffic indefinitely. "Silent drop → hang" becomes "unbounded growth → node OOM", triggered by an authorized peer.

I don't feel sam-node should own a JSON-RPC multiplexer. Stdio MCP is single-session by spec; no SDK multiplexes it, because there's no correct way to do so (server-initiated requests like sampling or elicitation have no attributable destination on a shared process). Any mux here — the current bridge, this queue, or a smarter id-rewriting version I prototyped — is protocol-level code we'd maintain and get subtly wrong over years. We've been there in Kubernetes.

What I'd like to do instead: use `mcp.CommandTransport` from the go-sdk we already depend on — one child process per session, exactly mirroring what backendTransport() already does for URL backends with StreamableClientTransport:

```go
case *api.RegisterServiceRequest_Command:
    cmd := exec.Command(x.Command.Command[0], x.Command.Command[1:]...)
    // env setup as today
    return &mcp.CommandTransport{Command: cmd}, nil
```

`Probe`, `Tools` and `HandleStreamPassThrough` only see an `mcp.Transport`, so they don't change. `StdioBridge` and `bridgeTransport` get deleted. Each peer gets its own process: no shared id space, no shared buffer, no cross-peer state or failure — which is what the stdio protocol assumes and what our Zero Trust guideline asks for.

Two things to keep in mind while doing it:

* `StdioBridge.ServeHTTP` (the SSE+POST handler the ingress serves for command backends) loses its home. Let's handle that separately; the streams path is what matters here.
* Slow-starting backends (npx …) will pay startup per session but we can optimize later if needed.

Would you be up for reworking this PR along those lines, keeping your subprocess-based regression tests adapted to the new transport? If you'd rather not, no problem — say so and I'll open it myself.

internal/node/stdio_bridge.go
fer-marino: Agreed, and you found a worse bug than the one I was chasing. Replied in full below - subscriberQueue is gone, this thread is moot now.

fer-marino:
@aojea You're right, and the id-collision you found is worse than what I was chasing - a shared process silently handing session A session B's tools/call result is a real correctness bug, not just a liveness one. Went with reworking it rather than leaving it to you, since it's a fairly contained change once you see it your way.

Pushed the rework (commit 82a6a86): `StdioBridge` and `bridgeTransport` are deleted. `MCPService.backendTransport` now builds a fresh `mcp.CommandTransport` per call for command backends - one subprocess per session, matching what it already did for URL backends via `StreamableClientTransport`, exactly as you sketched. `Probe`, `Tools` and `HandleStreamPassThrough` didn't need to change.

Two notes on the parts you flagged:

- `StdioBridge.ServeHTTP`'s local SSE/POST ingress route for command backends is gone with it - `baseService.Init` no longer builds a handler for them, so `Handler()` returns nil and the existing nil-checks in the ingress path 404 instead of crashing, per your "handle that separately." The one test that exercised that route (`TestStdioDatapathIntegration`) is removed, with a comment in its place pointing at this thread for why.
- Slow-starting backends now pay startup per session, as you noted. Not optimized in this PR.

New tests in `mcp_service_command_test.go` replace the old subscriber-queue ones: two concurrent command-backend sessions, both sending `id: 1` the way independent go-sdk clients naturally would, asserting each only ever sees its own subprocess's replies (tagged by PID). That's the direct regression test for the bug you found - it fails against the old shared-bridge design and passes now, by construction rather than by getting a shared id space right.

PR title/description updated to match. Full `internal/node` suite still passes (same five pre-existing, unrelated Windows-environment failures as before - none of them touch this code).

aojea:
failures look related

> not ok 12 Datapath: HTTP and Stdio services are reachable across nodes

internal/node/mcp_service.go
```diff
@@ -75,18 +77,43 @@ func (m *MCPService) Teardown() error {
...
+// what backendTransport already does for URL backends.
+//
+// The cost is a fresh process per Probe/Tools call and per mesh stream
+// instead of one long-lived one; command backends with slow startup pay
+// that repeatedly. Command backends also lose the local SSE/POST HTTP
+// ingress route StdioBridge.ServeHTTP used to provide (baseService.Init
+// no longer builds a handler for them) - the mesh stream path
+// (HandleStreamPassThrough) is what actually matters and is unaffected;
+// restoring a local-ingress equivalent is left for later, not folded into
+// this change.
```
aojea: try to tell the AI agent to not create this very large and verbose  comments, I like the comments but this is a book :)

fer-marino: Fair, noted - trimmed that one and a couple others down in 0f40c80.

fer-marino:
Yep - deleting StdioBridge.ServeHTTP along with bridgeTransport broke that route for real, not just hypothetically (the bats test proved it). Fixed in 0f40c80: StdioBridge is back, but trimmed to only what ServeHTTP needs. That route was never part of the id-collision bug - it already used a fresh channel per POST, no shared connection-wide id space. bridgeTransport/Subscribe() are still gone; MCPService.backendTransport is unchanged from 82a6a86.

fer-marino:
Fixed in ac74055, per your suggested check: `x.Command == nil || len(x.Command.Command) == 0` before indexing into it, with a regression test (nil CommandBackend and empty command slice, both asserting an error instead of a panic). All checks are green.

aojea:
/gemini review

added 3 commits for completeness and preserved the previous

aojea:
once CI passes I merge, thanks @fer-marino

aojea:
> not ok 1 a2a: stock SDK client chats with a mesh-hosted agent via the regenerated card

investigating flake, but seems unrelated to the PR

PR #390, opened 2026-09-12
use canonical peer ID string in revocation cache

IbrahimAhmed8:
This addresses a canonicalisation flaw in the revocation cache introduced in PR #356
The configuration loop and the event handler currently write the raw peer ID string to the cache. The problem is that enforcement checks always query this cache using the canonical base58 format.
If a node enrols using an alternative format like CIDv1 or base32 the system caches that exact raw string. When the enforcement logic later checks the base58 version it results in a cache miss and the banned node is granted access

The patch updates both write paths to store the decoded and normalised string, this aligns the stored keys with the lookup queries and ensures the bans are enforced correctly


aojea:
> If a node enrols using an alternative format like CIDv1 or base32 the system caches that exact raw string. When the enforcement logic later checks the base58 version it results in a cache miss and the banned node is granted access

can you add a regression test so we ensure we never regress?

aojea:
Also it seems we need to follow up, there are more places with this problem in the control plane: in each handler, canonicalize once right after peer.Decode (canonical := pID.String()) and use it for every store.* call and PublishEvent. 

For /revoke and HandleUserRevoke, decode the incoming ID and publish node.PeerID from the record. Keep the PoP challenge (EnrollChallenge(req.PeerId, …)) on the raw string the client actually signed, or switch it to canonical — our clients always send Host.ID().String(), so either works; canonical seems simpler and more consistent.

For discovery: store claimed.String() in Provider.PeerID and entryKey.

IbrahimAhmed8:
I will add the regression test shortly 
Applying this to the control plane and discovery handlers makes sense but canonicalising the database reads without migrating the existing rows will unban any legacy non canonical records. I'll put the migration and the store changes in a separate pr

aojea:
I expect something that covers more codepaths, like an integration test of a node enrolling with another id format or something

IbrahimAhmed8:
> I expect something that covers more codepaths, like an integration test of a node enrolling with another id format or something

ive also added a test covering initialisation

aojea:
/gemini review

ok, let's try to get this in and follow up on the pending things

PR #392, opened 2026-09-14
Add node self-reported service catalog and console mnemonics for node labels

fer-marino:
## Why

The admin console has no service-level visibility at all today: an operator can see enrolled nodes and routers, but has no way to answer "what services is this mesh actually running, and where". Separately, both the Nodes table and (now) the new Services table only ever show a bare peer ID, which is unreadable at a glance - there is no node mnemonic/name anywhere in SAM.

`MeshAdapter.DiscoverServices`/`GetNodeStatus` are real interface methods scaffolded for exactly the first problem, but both `NopMeshAdapter` and `P2PMeshAdapter` only `return nil, nil` and neither is called from anywhere. A faithful implementation would require the control plane to become a DHT participant itself (it currently isn't), which is a much bigger change than this gap warrants.

## What this does

**Push-based service catalog.** A node already knows its own local service list (`ListLocalServices`) and already runs a periodic control-plane check-in (`startPolicySyncLoop`). This adds a sibling loop (`CatalogReportInterval`, default 1m) that POSTs that list to a new `POST /nodes/catalog` endpoint, authenticated with the node's own biscuit so a node can only ever report on itself (peer ID is extracted server-side from the verified token, never trusted from the request body). The control plane caches the result in memory, keyed by peer ID - this is a live-status view, not authoritative state, so it's fine to lose on restart; the node re-reports on its next tick.

**Console "Services" view.** Joins the cached per-node catalog into a flat table (service name, type, description, reporting node, last-reported timestamp), using the existing `HandleAdminStatus` JSON response as the transport (no proto/wire changes - this is plain JSON like `/admin/bootstrap-tokens` already is).

**Node labels as mnemonics.** `EnrolledNode.Labels` already exists and is already populated from `sam-node.yaml`'s `labels:` key, but was never surfaced anywhere. Both the Nodes table and the new Services table now show it as a small `key=value` line under the peer ID.

Labels are display-only. They carry no cryptographic proof and are not wired into `api.TargetFactRules` or any other authorization path, so they must never become a trust primitive - only ever an operator-facing lookup aid alongside the real identifier (the peer ID).

## Testing

`go build ./...` passes. `go test ./internal/controlplane/... ./internal/node/... ./internal/console/...` shows the same pre-existing/environmental failures as on main (unrelated to this change, e.g. TestNodeProactiveTokenRefresh, a handful of internal/node backend/socket tests) and no new failures. Manually verified end-to-end against a local sam-one instance with two nodes reporting real services (screenshots below).
sam-nodes-labels
sam-services


fer-marino:
Thanks for the review jimmy - addressed all four points in 632b8dd.

Filtered nil elements out of a catalog report's services array server-side before caching (HandleNodeCatalog), and added the matching defensive skip on the console side. Reported At in the Services table now uses toLocaleString(), consistent with the rest of the console. Added unit tests: TestHandleNodeCatalog (plus auth/admission/method-not-allowed/nil-filtering variants) in internal/controlplane/catalog_test.go, and TestReportNodeCatalog (plus HTTP error propagation) in internal/node/controlplane_test.go.

aojea:
> Thanks for the review jimmy

hahah

do we need to add something to `tests/ui` for better coverage?

fer-marino:
Good question - pushed 6188cca which does two things.

First, a real bug this made me look for: peerCell (the label line under a peer ID) was injecting an inline style= attribute via innerHTML, which would break under any style-src CSP without 'unsafe-inline' - exactly what tests/ui/console.spec.js's "no inline style attribute" test exists to prevent, just not caught there since that test only scans the served index.html, not app.js's generated markup. Moved it to a real class (.cell-subtext).

Second, added two Playwright tests: the Services view is reachable from the nav, and it's deep-linkable via the hash, both asserting the empty state renders without a JS error. I couldn't do more than that from the UI suite - a populated Services table needs a real node self-reporting, and this suite has no fixture for that (Nodes and Enrollments don't get that kind of coverage here either, same reason). The actual reporting path is already covered at the Go level from the earlier review round: TestHandleNodeCatalog in internal/controlplane and TestReportNodeCatalog in internal/node.

Verified all of this locally end to end rather than just by inspection: built sam-control-plane + sam-console from this branch, ran them against a local mock OIDC issuer, and the full tests/ui/console.spec.js suite is green (17/17, including the two new tests).

aojea:
> sam-control-plane + sam-console from this branch

this make things like this easier https://github.com/google/sam/pull/360 ;) just building that component gives you a standalone control plane with the console

fer-marino:
Ha, noticed that once I went looking - #360 is exactly why building just cmd/sam-control-plane and cmd/sam-console gave me a working standalone stack in two build commands. Nice payoff.

aojea:
@fer-marino I hope you don't mind I took over this PR in https://github.com/google/sam/pull/408 , I made sure your authored work remains , all credit to you thanks

aojea:
> @fer-marino I hope you don't mind I took over this PR in #408 , I made sure your authored work remains , all credit to you thanks

PR #394, opened 2026-09-14
fix peer ID aliasing bugs across control plane and discovery

IbrahimAhmed8:
The same peer can be represented by different valid ID formats. Some parts of the code were storing the value provided by the client while other parts were using the canonical base58 form. This could cause the same peer to be stored under different keys and in the case of revoked nodes allow a different representation of the same id to avoid the existing check.
Peer IDs are now converted to their canonical form after decoding before being stored or used in lookups

the control plane and discovery handlers now use the same representation and the node cache is updated in the same way. Tests were added for peers using different id formats


api/network.go
```diff
@@ -76,6 +77,14 @@ const (
 	TokenRefreshCheckInterval = 10 * time.Minute
 )
 
+func CanonicalPeerID(s string) (string, error) {
+	p, err := peer.Decode(s)
+	if err != nil {
+		return "", err
+	}
+	return p.String(), nil
+}
```
aojea: I'm not convinced this should be an API, I know some AI agents recommend this but is more a normalization/canonicalization than an API, WDYT? 
Most of the rest of the code is also decoding directly, that to me is a proof that is not the right abstracion

IbrahimAhmed8: I went back through the call sites and I agree it makes more sense to drop the export.
There are only four uses at the moment all added in this pr. api isn't really the right place for it, given that the package is otherwise mainly wire constants and protobuf types

I'll make the changes now

internal/controlplane/mesh.go
```diff
@@ -165,6 +165,17 @@ func (p *P2PMeshAdapter) PublishEvent(ctx context.Context, eventType api.MeshEve
...
+	if peerID == "" {
+		return ""
+	}
+	p, err := peer.Decode(peerID)
+	if err != nil {
+		return peerID
+	}
+	return p.String()
+}
+
```
aojea: this is a bit odd, I think you just need to inline the code and return an error if you can not decode, otherwise it will try to create the event anyway ... remove the helper and before the event try to decode and return the error if it fails

internal/controlplane/peer_id_canonicalization_test.go
```diff
@@ -0,0 +1,368 @@
...
+	alias := cidAlias(t, id)
+
+	if status := register(t, priv, id.String(), "alias-attacker"); status != http.StatusOK {
+		t.Fatalf("initial registration: got %d, want 200", status)
+	}
+	if _, err := store.GetNode(ctx, id.String()); err != nil {
+		t.Fatalf("node was not stored under its canonical id: %v", err)
+	}
+
+	revokeData, _ := proto.Marshal(&api.TokenRevokeRequest{PeerId: id.String()})
```
aojea: test can not shallow errors, I need to add that to gemini review

internal/controlplane/server.go
```diff
@@ -1471,8 +1475,15 @@ func (s *Server) HandleEnrollStatus(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	pID, err := peer.Decode(peerID)
+	if err != nil {
+		http.Error(w, "Unauthorized", http.StatusUnauthorized)
```
aojea:
```suggestion
		http.Error(w, "Unauthorized, Invalid Peer ID", http.StatusUnauthorized)
```

internal/storage/sql_store.go
```diff
@@ -113,9 +114,10 @@ func (s *SQLStore) isPostgres() bool {
 }
 
 type migration struct {
-	version  int
-	sqlite   []string
-	postgres []string
+	version     int
+	sqlite      []string
+	postgres    []string
+	dataMigrate func(ctx context.Context, s *SQLStore, tx *sql.Tx) error
```
aojea: at this point we are still in alpha and I think we can make this breaking change, I do not want to add data migration of a database as a drive by PR, just remove this part in storage and we tackle it independently with a better battery of tests ... this is a critical part that require its own design

aojea:
few comments and almost there

/gemini review

internal/controlplane/mesh.go
IbrahimAhmed8: I'll sort the test errors and update that http message 
We have a slight snag with inlining the decode in the mesh package though. The decode function fails on empty strings so returning an error there will quietly break global events like key rotations which intentionally pass an empty peer ID. should I add a quick exception for empty strings before the decode or handle it differently?

aojea: ok, fair, then make it part of the function logic and add a comment explaining it, "events can have empty peer id",  but don't encapsulate the logic into this function, it will mislead people since the semantics are confusing

internal/storage/sql_store.go
aojea: @IbrahimAhmed8 you marked this as resolved but code still remains

IbrahimAhmed8: sorry I mistakenly resolved it bcause I thought I had already removed the remaining migration scaffolding while I was working through the other changes but ive removed it now

aojea:
Thanks

PR #395, opened 2026-09-14
discovery: record the canonical peer ID from an announcement

HosniBelfeki:
Follow-up to #390, one layer up.

## The gap

`Discovery.observe` decodes the announced peer ID to check it against the pubsub signer, then stores the **announced** string rather than the decoded one:

```go
claimed, err := peer.Decode(ann.GetPeerId())
if err != nil || claimed != msg.GetFrom() { ...drop... }
...
entryKey := ann.GetPeerId() + "|" + typeStr + "|" + ann.GetServiceName()
d.providers[entryKey] = Provider{ PeerID: ann.GetPeerId(), ... }
```

A peer ID has several valid encodings that all decode to the same peer, so a peer announcing itself in any of them passes the signer check and is recorded under that spelling. Measured on one generated key:

| announced | decodes to the signer | differs from `p.String()` |
|---|---|---|
| `12D3KooWEq4cBRc1RngNA1fm1HUzLSTU4uaZKcP3FcN1jdrGCdt5` (canonical) | ✓ | — |
| `bafzaajaiaejcastyfcoumkmw5vrzpbycgxfbuuafgyjtd2xolwtizcu7odcizi7k` (base32) | ✓ | ✓ |
| `k51qzi5uqu5di1eoj9gy72o724vd9pf8lsbrvjxyezl0j40twtgn54unf17obu` (base36) | ✓ | ✓ |
| `z5AanNVJCxnKr5h4A5iUoNm9Yb62xw8hkBfTpA6U74RcCB5yTR7bL5f` (base58btc) | ✓ | ✓ |

## Why it matters

Everything reading the view compares against a canonical `peer.ID.String()`. `PeerLabels` matches on equality, and `Provider.PeerID` reaches the node as `modelProvider.peerID`, which the inference facade passes straight to the revocation cache:

```go
isRevoked: func(peerID string) bool {
    return node.revokedPeers != nil && node.revokedPeers.Contains(peerID)
}
```

After #390 that cache is keyed canonically, so a peer stored under another spelling is one those lookups cannot find — including the facade's revocation check, which is exactly where a banned peer has a motive to pick a different spelling.

**Scoped honestly:** the connection gater still refuses the peer on the decoded `peer.ID`, and `handleBannedEvent` closes the connection, so this is a layer of defence that quietly stops working rather than a way through. It is also not reachable from an ordinary node — the announcer publishes `d.self.String()`, already canonical — so it needs a peer that deliberately announces a different spelling.

Independently of revocation, one peer can hold two view entries, one per spelling, against the `maxProviders` bound.

## The change

Two lines. The decoded value is already in hand from the signer check, so this only stops discarding it — same shape as the fix in #390, and the same root cause as the one it fixed: decode to validate, then keep the unvalidated form.

## Tests

`TestObserveCanonicalisesAnnouncedPeerID` announces each CIDv1 spelling and asserts the view records the canonical form and that `PeerLabels(canonical)` finds it, plus that two spellings of one peer do not accumulate two entries.

It first asserts each spelling really does differ from the canonical form **and** decodes back to the same peer, so if that ever stops being true the test fails loudly instead of passing for the wrong reason.

All four cases fail without the change:

```
--- FAIL: TestObserveCanonicalisesAnnouncedPeerID/base32
--- FAIL: TestObserveCanonicalisesAnnouncedPeerID/base36
--- FAIL: TestObserveCanonicalisesAnnouncedPeerID/base58btc
--- FAIL: TestObserveCanonicalisesAnnouncedPeerID/spellings_do_not_accumulate_entries
```

```
$ go build ./... && go vet ./internal/... ./api/...
$ go test ./internal/node/discovery/ -count=1
ok      github.com/google/sam/internal/node/discovery  2.531s
```

## Local verification caveat

Six tests in `internal/node` fail on my Windows workstation with and without this change — the Unix-socket and subprocess-backend ones. Same set on a clean checkout of the merge base. `make lint` and `make e2e-test` need a Linux toolchain I don't have locally, so CI is the authority.


aojea:
this is a dupe of https://github.com/google/sam/pull/394/ , let's work on a single place

aojea:
Fixed in https://github.com/google/sam/pull/394 so I'm going to close, anyway I want to thank you for keep contributing, just this time both of you collided :)

HosniBelfeki:
No problem at all, @aojea  Great minds think alike. :) I completely agree with keeping the work centralized in #394. Thanks for the quick review and the kind words I'm glad the issue is being resolved and I look forward to contributing again soon!

PR #415, opened 2026-09-17
Security review remediation: proof of possession, identity bans, type-bound dispatch, gossip validation, control-plane transport trust, secrets hygiene, mobile tokens

aojea:
Remediation of the 2026-09-16 security review, stacked as one commit per area so each can be read on its own. Everything below is alpha-breaking where noted.

## What changes

**Enrollment proves key possession (`fe9a110`).** `POST /register`, `POST /enroll` and `POST /routers/lease` now carry a timestamp and a signature over an endpoint-specific challenge made with the key being enrolled. Previously anyone could register an arbitrary peer id, and any enrolled node could rewrite a router's addresses with the router's own (widely shared) biscuit. Router leases are verified against the router's stored key and require the router role.

**Pre-authentication work is bounded (`0f2ceba`).** Inbound biscuits with appended blocks are refused before any Datalog runs; the authorizer has fact and iteration caps; auth-handshake streams get a per-peer rate limit and a deadline on both node and router.

**Control plane: identity bans and token spending (`cbd1e49`).** A banned OIDC identity can no longer mint bootstrap tokens or re-enroll a new device via `/user/*`. Bootstrap-token usage is consumed atomically (`UPDATE … WHERE usages < max`), the node record is written before an enrollment request is resolved, approve re-checks the token, and there is an unban path. IdP `roles` claims become `idp_role` facts. User tokens are clamped to 7 days / 10 uses; `/user/status` no longer returns the mesh policy to every user; the CLI ban path canonicalises peer ids and errors on zero rows.

**Node dispatch is bound to the service type (`3bccddc`).** A stream authorized for `mcp://foo` cannot reach `inference://foo`; the registry refuses one name under two types; ingress and egress reject `..` segments. Command-backed MCP services get a per-request id space (two callers with the same JSON-RPC id no longer receive each other's replies) and the SSE broadcast side is removed. The sidecar strips its own token from `Authorization` before forwarding (only when it equals the sidecar token; upstream credentials pass through). The reverse proxy addresses the backend by its configured host and drops inbound `X-Forwarded-*`. Remote `get_mesh_info` returns only peer id and DHT size.

**Gossip, relay and callee verification (`b0c5c3c`).** A GossipSub validator on the control-plane events topic rejects unsigned or forged events at the first hop, so a junk flood via the router can no longer exhaust the rate budget that real ban and rotation events share; the limit is keyed on the author. The relay ACL requires the *source* to be authenticated too. Every provider a node talks to must present a control-plane-signed biscuit bound to its peer id and carrying `role(node)` — on MCP streams, the OpenAI facade and the raw egress proxy — not only when labels or an egress floor apply. The discovery table caps entries per signer. The `mesh_pubsub_broadcast` / `poll_messages` / `subscribe_topic` MCP tools are removed; A2A is the agent-to-agent channel.

**Control-plane transport trust (`4d92e21`).** Whoever answers the control-plane URL is the trust root, so a plaintext `http://` URL to a non-loopback host is refused by `sam-node` and `sam-router` unless `--insecure-control-plane` is passed (charts, k8s templates, e2e and docs opt in for the in-cluster Service URL). `GET /keys` is signed by every valid key and accepted only when a key the receiver already trusts signed it. `KEY_ROTATION` events are signed by the retiring key — they were signed by the new key, which no node trusted yet, so rotation announcements were being dropped. A refreshed biscuit is verified (trusted signer, own peer id, role) before it replaces the identity, and a `403` on `/refresh` no longer `os.Exit`s the process.

**Secrets out of the tree and out of env (`0244e42`).** The docs-site demo recording carried a real daemon token (redacted; `hack/verify-secrets.sh` now fails CI on bearer tokens and PEM keys in tracked files). The mock OIDC issuer's RSA private key was committed three times; all three generate it at start-up. The sam-node chart's `apiToken: devtoken` env var becomes a generated Secret mounted as a file; the router bootstrap token is one use per replica instead of 999999; the sam-node Deployment and the internal postgres get security contexts and `automountServiceAccountToken: false`. The agent skill no longer tells the model to read MCP client config files to recover the node token. The dead TLS-MITM gateway in `internal/sambox` is deleted.

**Mobile (`3b6dc32`).** The app's sidecar token was the fixed string `secret-token`, and Android loopback is shared by every installed app; it is now generated per device and shown on the Config tab. The phone-sensors MCP backend was unauthenticated on `127.0.0.1:9090`; it now binds a random port with a per-launch bearer token. Location is coarse-only (network provider, ~1 km) to match what the switch promises; `ACCESS_FINE_LOCATION` is dropped; `allowBackup=false`. Supporting this, a service may name a credential file in `target_auth_path`; a credential written into `target_url` in `sam-node.yaml` is refused.

**Hygiene (`024de4c`).** Admin token masked in the deploy workflow and minted over a port-forward; SQLite files 0600; secrets read from files unset their env var so subprocesses don't inherit them; response bodies from the control plane are size-capped; log lines are bounded; `.dockerignore`; release workflow limited to `v*` tags without `packages: write`; dependabot covers the other ecosystems.

## Breaking

- Clients must send the challenge fields on `/register`, `/enroll`, `/routers/lease`.
- `/keys` responses are signed; unsigned ones are refused. Nodes/routers older than this cannot follow a rotation.
- Plaintext `http://` to a non-loopback control plane needs `--insecure-control-plane`.
- Providers must carry `role(node)`.
- The three custom pubsub MCP tools are gone.
- sam-node chart: `apiToken` defaults to a generated Secret; `--api-token-path` replaces the env var.

## Deferred (reasons in the commit messages)

Gossip/DHT peer admission filters (would break discovery for peers that have not handshaken), the `/debug/connect-peer` dial oracle, a per-route control-plane rate limiter, cosign signing of releases (needs a new Action), a release-signing keystore for the APK, and the hub canaries' `SAM_API_TOKEN` env vars and floating images.

## Operator action

Rotate the daemon token on the machine that recorded `site/static/demo.cast` (`rm ~/.config/sam-mesh/api-token && sam-node run --daemonize`).

## Verification

`make build`; golangci-lint (root and `cmd/nano-init`) 0 issues; deadcode; `hack/verify-generated.sh`; `hack/verify-secrets.sh`; helm unittest 46/46; `flutter analyze`; full `go test -race ./...` including `tests/integration`. Every security branch has a negative test; the gossip-flood and provider-role tests were checked to fail with the fix reverted. Four independent review passes over the final diff; their findings (provider role, explicit StrictSign, loopback edge cases) are folded in.


aojea:
Follow-ups pushed:

- **CI (kind-mesh e2e, bats e2e)** — `843a4b8`: the internal postgres moved to `runAsUser: 70` in this PR, but with the PVC mounted at `/var/lib/postgresql/data` `initdb` could not chmod the root-owned mount root and crash-looped, so the control plane never answered `/info` and the bootstrap hook timed out. `PGDATA` now points at the `pgdata/` subdirectory postgres creates and owns (same shape as `.github/k8s/sam-control-plane-template.yaml`). Reproduced in docker and on kind: pod crash-loops without it, Ready in ~10s as uid 70 with it. Upgrade note added to `values.yaml`.
- **StdioBridge lock split** (review comment) — `b191901`: legitimate and worse than a stall — with a subprocess it is a deadlock (backend blocked writing stdout → not reading stdin → caller blocked in the stdin write holding the lock `deliver` needs to drain stdout). Writes now serialize on their own mutex; `mu` is never held across I/O. New `TestStdioBridge_BlockedStdinWriteDoesNotStallDelivery` fails on the previous code.
- **Package-level `allowInsecureControlPlane`** (review comment) — acknowledged, not changed here. The policy is deliberately process-wide: the control-plane URL enters from a flag, the stored config, and the mobile FFI, and every request has to be held to the same rule regardless of entry point; `internal/node` tests don't use `t.Parallel`. Encapsulating the ~8 package-level control-plane calls into a client struct is a reasonable refactor for a follow-up.

aojea:
kind-mesh e2e: `development/kind/run-local-node.sh` builds its control-plane URL from the Gateway's LoadBalancer IP at runtime, so it was the one plaintext launch site the flag sweep missed. Opted in with `--insecure-control-plane` (the docker bridge on the dev machine is the trust boundary; the script already fetches the admin token over that hop). TLS for the kind flow would need an HTTPS Gateway listener, a TLS mode on the control plane and a CA-pinning flag on sam-node — a separate change, not dev-only plumbing to bolt on here.

aojea:
**Bats e2e triage** (run 35217706060, 5 failures) — each failure was classified as *breaking change*, *test relying on an assumption this PR removes*, or *regression*, before touching anything:

| # | Test | Verdict | Fix |
|---|------|---------|-----|
| 4, 6 | Auth Flow 2 (device) / Flow 4 (bootstrap token) | **Breaking change (intended).** `sam-node join` now refuses `http://` to a non-loopback control plane. These two call sites pass the URL positionally and were missed when the other plaintext call sites opted in. | 77379f3: add `--insecure-control-plane` to the two `join` invocations. |
| 12 | Datapath: stdio service reachable | **Wrong assumption.** The test expected the POST reply to be broadcast on a shared SSE stream (the M18 finding this PR fixes). MCP Streamable HTTP lets a server answer GET with 405 when it offers no server-push stream. | 77379f3: two concurrent POSTs with the same JSON-RPC id must each get their own reply; GET must return 405. |
| 24, 37 | Relay discovery / sam-one dataplane | **Regression from H3-a.** Unconditional callee verification made the label gate the first dial to a peer, with a plain context: libp2p then demands a direct connection, and relay-only or not-yet-resolved peers fail with `no addresses` → egress 403. Never surfaced before because those flows have no labels/floor. | d6619e1: gate runs `preparePeerAddrs` and dials with `WithAllowLimitedConn`, same as the guarded call; unit test pins the opt-in. |

Verified locally with `E2E_REUSE_CLUSTER=1 bats tests/e2e/{datapath,relay,auth_flows}.bats` → 7/7, plus `standalone.bats` (sam-one) → ok. Lint clean, `go test -race ./internal/node/` green.

PR #428, opened 2026-09-17
Close latent hole in service re-declare path

IbrahimAhmed8:
In `internal/node/service_registry.go` registering a service again under the same name and type overwrites the map entry but fails to call Teardown on the replaced instance. This closes a latent hole in the re-declare path rather than an observed leak, since nothing in production currently re-registers a command backend. Register now captures the replaced entry during the map swap and explicitly tears it down outside the lock

The test in `internal/node/service_registry_test.go` verifies the replaced instance receives a `teardown` call


aojea:
The replaced-instance leak is real and worth fixing, but (a) the mobile retry loop can't trigger it — MobileService only supports targetUrl, and Teardown is a no-op for URL backends; nothing in production re-registers a command backend today — and (b) the pending/generation machinery guards races that have no callers (Unregister is test-only, registration is single-goroutine). Please reduce this to tearing down the replaced entry after the swap (~6 lines), keep TestServiceRegistry_ReplacingServiceTearsDownPrevious, and drop the subprocess test and the concurrency scaffolding. Fix the description to say it's closing a latent hole in the re-declare path rather than an observed leak.

```go
r.mu.Lock()
replaced := r.services[info.Name]
r.services[info.Name] = svc
r.mu.Unlock()
if replaced != nil && replaced != svc {
    if err := replaced.Teardown(); err != nil {
        logger.Errorf("[ServiceRegistry] Teardown replaced %s: %v", info.Name, err)
    }
}
```

IbrahimAhmed8:
> The replaced-instance leak is real and worth fixing, but (a) the mobile retry loop can't trigger it — MobileService only supports targetUrl, and Teardown is a no-op for URL backends; nothing in production re-registers a command backend today — and (b) the pending/generation machinery guards races that have no callers (Unregister is test-only, registration is single-goroutine). Please reduce this to tearing down the replaced entry after the swap (~6 lines), keep TestServiceRegistry_ReplacingServiceTearsDownPrevious, and drop the subprocess test and the concurrency scaffolding. Fix the description to say it's closing a latent hole in the re-declare path rather than an observed leak.
> 
> ```go
> r.mu.Lock()
> replaced := r.services[info.Name]
> r.services[info.Name] = svc
> r.mu.Unlock()
> if replaced != nil && replaced != svc {
>     if err := replaced.Teardown(); err != nil {
>         logger.Errorf("[ServiceRegistry] Teardown replaced %s: %v", info.Name, err)
>     }
> }
> ```

Fair enough. I stripped the scaffolding
Thank you

aojea:
Thanks

PR #452, opened 2026-09-19
ci(flutter): boot the cached emulator snapshot with a real readiness check

aojea:
## What

Fixes the recurring `e2e-android` failures in [test_flutter.yaml](https://github.com/google/sam/actions/workflows/test_flutter.yaml?query=is%3Afailure): 5 of the last 30 runs, all with the same signature and all before `mobile_e2e.sh test` even started. Keeps the cached AVD snapshot, so the job stays at ~7 minutes.

## Why

Every failure looks like this, inside the `android-emulator-runner` action, right after it restores the cached snapshot:

```
Successfully loaded snapshot 'default_boot' using 1517 ms
adb -s emulator-5554 shell getprop sys.boot_completed   -> adb: device offline
adb -s emulator-5554 shell getprop sys.boot_completed   -> 1   "Emulator booted."
adb -s emulator-5554 shell input keyevent 82
java.lang.IllegalStateException: android.os.ServiceManager$ServiceNotFoundException: No service published for: input
```

A restored guest already has `sys.boot_completed=1` from the saved state, so the action's boot wait returns two seconds after the load, as soon as adb reconnects, and its unconditional `input keyevent 82` races the system services coming back up. The action offers no way to skip or retry that command ([`launchEmulator` in v2.38.0](https://github.com/ReactiveCircus/android-emulator-runner/blob/v2.38.0/src/emulator-manager.ts), unchanged on `main`).

## Fix

The action is now used only where it is not racy: creating the AVD and its snapshot on a cache miss. The test step boots the snapshot from a `run:` step that does what the action did (same `sdkmanager` installs, same emulator flags, same animation settings), but waits until `input keyevent 82` itself succeeds before declaring the emulator ready. That is the exact command that was failing, so the readiness signal cannot be wrong in the same way again. `adb emu kill` runs in an `if: always()` step.

The first commit on this branch was a cold-boot fix; the second replaces it to keep the snapshot's ~5 min saving. Squash on merge.


.github/workflows/test_flutter.yaml
```diff
@@ -114,20 +114,43 @@ jobs:
...
+    # input"), failing one run in six before the test starts. Booting here lets
+    # the readiness check wait for the thing that actually failed.
+    - name: Boot the emulator from the cached snapshot
+      run: |
+        set -o errexit -o nounset -o pipefail
+        export PATH="$ANDROID_HOME/cmdline-tools/latest/bin:$ANDROID_HOME/platform-tools:$ANDROID_HOME/emulator:$PATH"
+        echo "$ANDROID_HOME/platform-tools" >> "$GITHUB_PATH"
+        echo "ANDROID_AVD_HOME=$HOME/.android/avd" >> "$GITHUB_ENV"
+        export ANDROID_AVD_HOME="$HOME/.android/avd"
+        yes | sdkmanager --licenses >/dev/null
```
IbrahimAhmed8: could we append `|| true` to the `sdkmanager` line to stop the pipefail abort or perhaps keep the original action and pass `-no-snapshot-load` to avoid the stale state?

